PR #11, opened 2020-04-27
US-ShortMessage-02 : Display 'Switch to an article' when the number of caracters exceeds 1300

Rdidmohamedelhadi:


webapp/src/main/webapp/js/newsComposerPlugin.js
```diff
@@ -25,5 +25,28 @@
     }
   };
 
+  const hintActivityComposerPlugin = {
```
thomasdelhomenie: Your latest commit should be reverted IMO. We are in the News project, so it is better to name the plugin with a meaningful name, not a generic one.

thomasdelhomenie (changes requested):

webapp/src/main/webapp/js/newsComposerPlugin.js
Rdidmohamedelhadi: ok

PR #39, opened 2020-09-09
MAINT-27951_MAINT-27924: Open internal links in the same browser tab and the external ones in a news browser tab

azayati:
- Internal links inside or outside the news body should be opened in the same browser tab
- External links inside the news body should be opened in a new browser tab


azayati:
Backport [commit](https://github.com/exo-addons/news/commit/6fc2a1a5b7039b24df82930773bb5b281b706619)

mzorai (approved):

PR #46, opened 2020-09-17
MAINT-27974: Not allow news activities deletion

azayati:
News and shared news activities deletion should no more be allowed since it makes them not accessible from news application page


azayati:
Backport [commit](https://github.com/exo-addons/news/commit/ba3d6b0aa28a2785c3863dc2c623080014602d7f)

ahamdi (approved):

PR #56, opened 2020-12-22
MAINT-38335 : Mentioning users in an article is not working well

Rdidmohamedelhadi:
**The problem** when mentioning user at the start of a sentence the selected users is not displayed well.
**So** to insure that all mentioning user are treated I add some codes in `formatMention()` 
**The second** problem in `CK` doesn't detect the mentioning users at the end of the sentence **so** when I post article we read the last value in the `CK`

### DIFF
--- a/services/src/main/java/org/exoplatform/news/NewsServiceImpl.java
+++ b/services/src/main/java/org/exoplatform/news/NewsServiceImpl.java
@@ -1203,11 +1203,11 @@ public class NewsServiceImpl implements NewsService {
       String next = (String) tokenizer.nextElement();
       if (next.length() == 0) {
         continue;
-      } else if (next.charAt(0) == '@') {
-        String username = next.substring(1);
+      } else if (next.contains("@")) {
+        String username = next.split("@")[1];
         org.exoplatform.social.core.identity.model.Identity identity = loadUser(username);
         if (identity != null) {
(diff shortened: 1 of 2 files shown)


services/src/main/java/org/exoplatform/news/NewsServiceImpl.java
```diff
@@ -1210,6 +1210,20 @@ public String formatMention(String text) {
...
+          next = "<p><a href=\"" + CommonsUtils.getCurrentDomain() + identity.getProfile().getUrl() + "\">"
+                  + encoder.encodeHTML(identity.getProfile().getFullName()) + "</a>";
+        }
+      } else if (next.contains("@")) {
+        String username = next.split("@")[1];
+        org.exoplatform.social.core.identity.model.Identity identity = loadUser(username);
+        if (identity != null) {
+          next = next.split("@")[0] + "<a href=\"" + CommonsUtils.getCurrentDomain() + identity.getProfile().getUrl() + "\">"
+                  + encoder.encodeHTML(identity.getProfile().getFullName()) + "</a>";
+        }
```
ahamdi: Could we combine the 3 checks in one and just remove all characters before the @ ? we can imagine having other HTML tags before the @ character

ahamdi (changes requested):

ahamdi (approved):
LGTM :+1:

PR #62, opened 2021-01-26
Task-39856 : XSS issue in news content

rdenarie:
Before this fix a stored XSS issue can occur in new content.
This fix sanitize the html content before displaying it.


boubaker:
could you add UT please to ensure that this fixes bug and especially to ensure that future maintenance/improvements doesn't break this new logic.

boubaker (approved):

PR #69, opened 2021-02-03
TASK-39258: Add notification when being mentioned in an article

aelbenney:

PR #76, opened 2021-02-16
TASK-39258: Add notification when being mentioned in an article (#69)

mzorai:


ahamdi (changes requested):
This should be refactored to follow the PR proposed here https://github.com/exoplatform/news/pull/73

mzorai:
> This should be refactored to follow the PR proposed here #73

cc @aelbenney could you please help in this

aelbenney (dismissed):

services/src/main/java/org/exoplatform/news/NewsUtils.java
```diff
@@ -0,0 +1,103 @@
...
+
+public class NewsUtils {
+
+  /**
+   * Formats the body of the news to add profile link when mentioned
+   *
+   * @param news
+   * @return News with formatted body
+   */
+  public static News formatNews(News news) {
```
ahamdi: This function  is no more needed I think

ahamdi (changes requested):

ahamdi (approved):

PR #82, opened 2021-02-24
TASK-40065: Image inserted in news body are now displayed in the details view

NaderSai:
Problem: the source of the image in the sanitized body was different from the real source.
How it was solved: by applying the unescapeHtml method to recover the original source image.


lbengayed (approved):
LGTM

ahamdi (approved):

PR #85, opened 2021-02-25
TASK40065: Fix displaying images inserted in news body

NaderSai:
Problem: the source image in the getbody() method was the same like in the temporary file which is not the url read from the database.
How it was solved: get the url read from the database with applying replaceImagesURLs method to the getbody() result.


ahamdi:
Please add a description to the PR

ahamdi (approved):

PR #94, opened 2021-04-12
Task-42704: Enable post article on redactional space (space with redactors) when super manager or space manager

hassenelkamel-exo:
**Current behavior:** When adding at least a redactor for a space, the space manager or a super user or a space admin can no more post article.
**New Behavior:**  When adding at least a redactor for a space, the space manager or a super user or a space admin can post article.


azayati (dismissed):
LGTM after QA validation

azayati (approved):
LGTM

PR #103, opened 2021-04-30
Task-43425:  Delete an article from news application

hassenelkamel-exo:
Current behaviour:

- It is not possible to remove a news article from news application

New behaviour:

- It will be possible to remove a news article from news application with its related activities


azayati (dismissed):
LGTM after QA validation

azayati (approved):
LGTM

PR #114, opened 2021-05-27
TASK-44754: Make sure that the notification of a shared article in space after renaming the space redirects to the right url.

mdiha:
In the set up process of the space url inside the method getNotificationActivityLinkForNotSpaceMembers(), the groupId part is explicitly set to the space's pretty name instead of the space's groupId.
Thus when updating the space's name, the space's pretty name is also updated but for some reason the space's groupId is not updated so when returning the space's url after renaming a space the url becomes incoherent.
Also in the last part of the url the space's display name is used instead of the space's pretty name.
This fix makes sure to use the right attributes while setting up the space's url.


ahamdi (changes requested):
Please provide a description for this PR, and add a Unit test

ahamdi (dismissed):
LGTM

ahamdi (approved):

PR #117, opened 2021-05-28
Task-44606: Update newsComposerPlugin by adding enabled method to plugin params

AzmiTouil:
The problem is that the newsComposerPlugin is loaded only at the first site page access and not when switching from one page to another, so the "enabled" parameter will keep the initial value.
To fix this problem, we have to call the "enabled" method each time the component loads.

### DIFF
--- a/webapp/src/main/webapp/js/newsComposerPlugin.js
+++ b/webapp/src/main/webapp/js/newsComposerPlugin.js
@@ -5,7 +5,6 @@ const newsActivityComposerPlugin = {
   labelKey: 'news.composer.write',
   description: 'news.composer.write.description',
   iconClass: 'newsComposerIcon',
-  enabled: false,
   onExecute: function (attachments) {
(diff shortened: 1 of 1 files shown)


webapp/src/main/webapp/js/newsComposerPlugin.js
```diff
@@ -20,6 +19,11 @@ const newsActivityComposerPlugin = {
     }
 
     window.open(url, '_blank');
+  },
+  enabled: function () {
+    return  canUserCreateNews(eXo.env.portal.spaceId).then(canCreateNews => {
```
azayati:
```suggestion
return eXo.env.portal.spaceId &&  canUserCreateNews(eXo.env.portal.spaceId).then(canCreateNews => {
      return canUserCreateNews() == 'true'
     });
```

azayati (changes requested):
See review

PR #119, opened 2021-05-31
Taask-44804: Fix add/edit news rest service permissions

azayati:
Before this fix:

- It is not possible for super admin or super spaces admin who are not space members to create news (write an article not shown on composer) or edit news (from news detail and news app)
- It is possible for a redactional space members to create a news directly with the rest service
- It is possible for any space member to edit a news even if he is not the news poster directly with the rest service (It is not possible by interface since the edit button has the right

This PR will adjust and harmonize add/edit news rest service permissions with UI ones


hassenelkamel-exo (approved):
LGTM

PR #127, opened 2021-06-04
TASK-44216: Make sure that non-breaking spaces aren't considered in news's body.

mdiha:
In news's article body the non-breaking spaces are considered as plain text, i fixed the computed property in order to remove non-breaking spaces while checking for enabling the post button.

### DIFF
--- a/webapp/src/main/webapp/news-activity-composer-app/components/ExoNewsActivityComposer.vue
+++ b/webapp/src/main/webapp/news-activity-composer-app/components/ExoNewsActivityComposer.vue
@@ -261,7 +261,7 @@ export default {
       return this.newsId !== null;
     },
     postDisabled: function () {
-      return this.uploading || !this.news.title || !this.news.title.trim() || !this.news.body || !this.news.body.replace(/&nbsp;/g, '').trim();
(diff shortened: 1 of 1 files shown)


webapp/src/main/webapp/news-activity-composer-app/components/ExoNewsActivityComposer.vue
```diff
@@ -261,7 +261,7 @@ export default {
       return this.newsId !== null;
     },
     postDisabled: function () {
-      return this.uploading || !this.news.title || !this.news.title.trim() || !this.news.body || !this.news.body.replace(/&nbsp;/g, '').trim();
+      return this.uploading || !this.news.title || !this.news.title.trim() || !this.news.body || !this.news.body.replace(/<p>&nbsp;<\/p>/g, '').replace(/&nbsp;/g, '').trim();
```
ahamdi, line 264: Could you use https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/innerText to check the content of the ckeditor instance

mdiha: In order to use the innerText property we need to perform it on an element, but if we don't call the news.body inside the computed property then the changes of the article body won't be triggered.

ahamdi (approved):

PR #129, opened 2021-06-07
Task-44608: Add filter by drafts in news app

AzmiTouil:
Prior to this change, it is not possible to view my drafts news from news apps. Thus, we will implement a new filter by drafts in the news App.

### DIFF
--- a/services/src/main/java/org/exoplatform/news/queryBuilder/NewsQueryBuilder.java
+++ b/services/src/main/java/org/exoplatform/news/queryBuilder/NewsQueryBuilder.java
@@ -63,15 +63,11 @@ public class NewsQueryBuilder {
           sqlQuery.append("exo:author = '").append(filter.getAuthor()).append("' AND ");
         }
         if (filter.isDraftNews()) {
-          sqlQuery.append("publication:currentState = 'draft' AND jcr:path LIKE '/Groups/spaces/%' ORDER BY ")
-                  .append(filter.getOrder())
-                  .append(" DESC");
+          sqlQuery.append("publication:currentState = 'draft'");
         } else {
-          sqlQuery.append("publication:currentState = 'published' AND jcr:path LIKE '/Groups/spaces/%' ORDER BY ")
-                  .append(filter.getOrder())
-                  .append(" DESC");
+          sqlQuery.append("publication:currentState = 'published'");
         }
-
+        sqlQuery.append(" AND jcr:path LIKE '/Groups/spaces/%' ORDER BY ").append(filter.getOrder()).append(" DESC");
       } else {
         throw new Exception("Unable to build query, filter is null");
       }
--- a/services/src/main/java/org/exoplatform/news/queryBuilder/NewsQueryBuilder.java
+++ b/services/src/main/java/org/exoplatform/news/queryBuilder/NewsQueryBuilder.java
@@ -62,9 +62,16 @@ public class NewsQueryBuilder {
         if (StringUtils.isNotEmpty(filter.getAuthor())) {
           sqlQuery.append("exo:author = '").append(filter.getAuthor()).append("' AND ");
         }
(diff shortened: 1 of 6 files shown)


services/src/main/java/org/exoplatform/news/queryBuilder/NewsQueryBuilder.java
```diff
@@ -62,9 +62,16 @@ public StringBuilder buildQuery(NewsFilter filter) throws Exception {
         if (StringUtils.isNotEmpty(filter.getAuthor())) {
           sqlQuery.append("exo:author = '").append(filter.getAuthor()).append("' AND ");
         }
-        sqlQuery.append("publication:currentState = 'published' AND jcr:path LIKE '/Groups/spaces/%' ORDER BY ")
-                .append(filter.getOrder())
-                .append(" DESC");
+        if (filter.isDraftNews()) {
+          sqlQuery.append("publication:currentState = 'draft' AND jcr:path LIKE '/Groups/spaces/%' ORDER BY ")
```
azayati: you can put AND jcr:path LIKE '/Groups/spaces/%' ORDER BY ").append(filter.getOrder()).append(" DESC") outside the if/else block IMHO since it is common between the two conditions

webapp/src/main/webapp/news/components/NewsApp.vue
```diff
@@ -79,9 +80,10 @@
             <exo-news-details-action-menu
               v-if="news.canEdit "
               :news="news"
-              :show-edit-button="news.canEdit"
+              :show-edit-button="news.canEdit && !news.draft"
```
azayati, line 83: the edit is replaced by resume ?

webapp/src/main/webapp/news/components/NewsApp.vue
```diff
@@ -110,7 +112,7 @@
               <i class="uiIconClock"></i>
               <span>{{ news.creationDate }}</span>
             </div>
-            <div class="newsViews">
+            <div class="newsViews" v-if="!news.draft">
```
azayati, line 115: To be verified if we display that also for draft news

webapp/src/main/webapp/news/components/NewsApp.vue
```diff
@@ -287,7 +290,7 @@ export default {
           newsText: this.getNewsText(item.summary, item.body),
           illustrationURL: `${newsIllustration}?${newsIllustrationUpdatedTime}`,
           title: item.title,
-          creationDate: newsPublicationDate,
+          creationDate: item.publicationState !== 'draft' ? newsPublicationDate : newsUpdateDate,
```
azayati: to verify if it is not always update date

azayati (changes requested):
See review

services/src/main/java/org/exoplatform/news/queryBuilder/NewsQueryBuilder.java
AzmiTouil: Done

webapp/src/main/webapp/news/components/NewsApp.vue
AzmiTouil: Yes

webapp/src/main/webapp/news/components/NewsApp.vue
AzmiTouil: @samehaidi confirmed that we need to hide the views count for the draft

webapp/src/main/webapp/news/components/NewsApp.vue
AzmiTouil: Verified!

azayati: we can change this to updatedDate, it is more coherent IMHO

azayati (changes requested):
See review

azayati (dismissed):
LGTM after QA validation

azayati (approved):
LGTM

PR #132, opened 2021-06-08
45419 : Fix Tests

boubaker:


ahamdi (approved):

services/src/test/resources/conf/configuration.xml
```diff
@@ -1,69 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
 
 <configuration xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.exoplatform.org/xml/ns/kernel_1_2.xsd http://www.exoplatform.org/xml/ns/kernel_1_2.xsd" xmlns="http://www.exoplatform.org/xml/ns/kernel_1_2.xsd">
-  <component>
```
hbenali, line 4: This file is almost empty (semantically empty). Shouldn't we delete it?

boubaker: Yes, you 're right, it could be removed, but it's not a big deal since it's in test folder. Thanks for your remark @hbenali

PR #133, opened 2021-06-08
Task-45350: Implement the resume and delete actions for the draft news

AzmiTouil:
This PR will implement the resume and delete actions for the draft news displayed in the news app from the drafts filter

### DIFF
--- a/webapp/src/main/webapp/news/components/NewsApp.vue
+++ b/webapp/src/main/webapp/news/components/NewsApp.vue
@@ -67,11 +67,12 @@
         <a
           :href="news.url"
           :style="{ 'background-image': 'url(' + news.illustrationURL + ')' }"
-          class="newsSmallIllustration"></a>
+          class="newsSmallIllustration"
+          :target="news.draft ? '_blank' : '_self'"></a>
         <div class="newsItemContent">
           <div class="newsItemContentHeader">
             <h3>
-              <a :href="news.url">{{ news.title }} </a>
+              <a :href="news.url" :target="news.draft ? '_blank' : '_self'">{{ news.title }} </a>
             </h3>
             <news-spaces-shared-in
               v-if="news.activities && news.activities.split(';')[1]"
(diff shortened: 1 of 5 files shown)


webapp/src/main/webapp/news/components/NewsApp.vue
```diff
@@ -384,6 +385,10 @@ export default {
       const editUrl = `${eXo.env.portal.context}/${eXo.env.portal.portalName}/news/editor?spaceId=${news.spaceId}&newsId=${news.newsId}&activityId=${news.activityId}`;
       window.open(editUrl, '_blank');
     },
+    resume(news) {
```
azayati: I think that the resume action is the same as edit ?

azayati (changes requested):
See review

azayati (changes requested):
When we click on the draft news, it should be like a resume action

webapp/src/main/webapp/news/components/NewsApp.vue
```diff
@@ -119,7 +119,7 @@
             </div>
           </div>
           <div class="newsItemContentDetails">
-            <a :href="news.url">
+            <a :href="news.url" :target="news.draft ? '_blank' : '_self'">
```
azayati: I think that you can set news.target = news.draft ? '_blank' : '_self' when building the news list

webapp/src/main/webapp/news/components/NewsApp.vue
```diff
@@ -293,7 +294,7 @@ export default {
           updatedDate: item.publicationState !== 'draft' ? newsPublicationDate : newsUpdateDate,
           spaceDisplayName: item.spaceDisplayName,
           spaceUrl: item.spaceUrl,
-          url: item.url,
+          url: item.publicationState === 'draft' ? `${eXo.env.portal.context}/${eXo.env.portal.portalName}/news/editor?spaceId=${item.spaceId}&newsId=${item.id}` : item.url,
```
azayati, line 297: it not works if the keep the same url ?

azayati (changes requested):
See review

azayati (approved):
LGTM after QA validation

PR #136, opened 2021-06-09
TASK-44216: Make sure that non-breaking spaces aren't considered in news's body nor spaces.

mdiha:
I made sure to strip html tags for the computed property "postDisabled" in order to avoid enabling the post button for a body with white space. But in this case it won't be enabled for a body article with just an image or a video link, there must be some text.


webapp/src/main/webapp/news-activity-composer-app/components/ExoNewsActivityComposer.vue
```diff
@@ -261,7 +261,7 @@ export default {
       return this.newsId !== null;
     },
     postDisabled: function () {
-      return this.uploading || !this.news.title || !this.news.title.trim() || !this.news.body || !this.news.body.replace(/<p>&nbsp;<\/p>/g, '').replace(/&nbsp;/g, '').trim();
+      return this.uploading || !this.news.title || !this.news.title.trim() || !this.news.body || !this.news.body.replace(/<[^>]*>?/gm, '').replace(/&nbsp;/g, '').trim();
```
ahamdi: This regex may remove text even if it has just < charcter without an ending > charcter.
could you use https://developer.mozilla.org/en-US/docs/Web/API/DOMParser ?

ahamdi (changes requested):

webapp/src/main/webapp/news-activity-composer-app/components/ExoNewsActivityComposer.vue
mdiha: Okay

ahamdi (approved):
LGTM

PR #137, opened 2021-06-09
Task-45407: Fix news body and summary display in news app

AzmiTouil:
Prior to this change, we have a bad display in news app when a news body or news summary contains long words. This fix will break the long words for both body and summary.


azayati (approved):
LGTM after QA validation

PR #138, opened 2021-06-09
Task-43588: Fix horizontal scroll in snapshot page when a news with a long summary is pinned

hassenelkamel-exo:
**Actual behavior:**  When we pin a news having a long summary, an horizontal scroll coming from latest news widget will be shown on snapshot page
**New behavior:**  We add a word-break css style for body content and title in order to split the word (not in the middle) and wrap it into next line.


azayati (dismissed):
LGTM after QA validation

azayati (approved):
LGTM

PR #139, opened 2021-06-09
Task-44741: Delete the draft drawer from the creation news form

AzmiTouil:
This PR will remove the draft drawer in news creation form since we have added the drafts filter on news app.


azayati (changes requested):
I think that the method [getNewsDrafts ](https://github.com/exoplatform/news/blob/task-44741/services/src/main/java/org/exoplatform/news/NewsServiceImpl.java#L626 is nomore needed, to be deleted also IMHO

azayati (approved):
LGTM after QA validation

PR #142, opened 2021-06-11
Task-44749: Save draft while editing a posted article

AzmiTouil:
Prior to this change, it was not possible to save drafts when editing an already posted article. After this change, the posted news drafts are saved while editing and available for the news author from drafts filter of news application.


azayati (approved):
LGTM

PR #145, opened 2021-06-15
Task-45574: Delete posted article draft

AzmiTouil:
Prior to this change, it is not possible to delete a draft of a posted article. So, when deleting a draft of a posted article, we restore it to the last published version


services/src/main/java/org/exoplatform/news/NewsServiceImpl.java
```diff
@@ -102,6 +105,10 @@
 
   private final static String      PLATFORM_ADMINISTRATORS_GROUP   = "/platform/administrators";
 
+  public static final String       CURRENT_STATE                   = "publication:currentState";
```
azayati: I think that you can use StageAndVersionPublicationConstant.CURRENT_STATE instead

services/src/main/java/org/exoplatform/news/NewsServiceImpl.java
```diff
@@ -102,6 +105,10 @@
...
+
+  public final static String       DRAFT                           = "draft";
```
azayati: I think that you can use PublicationDefaultStates.DRAFT instead

services/src/main/java/org/exoplatform/news/NewsServiceImpl.java
```diff
@@ -637,6 +644,19 @@ public void deleteNews(String newsId) throws Exception {
...
+          if (node.hasProperty(CURRENT_STATE) && node.getProperty(CURRENT_STATE).getString().equals(DRAFT)) {
+            String nodeVersionUUID = (node.hasProperty(AuthoringPublicationConstant.LIVE_REVISION_PROP)) ?
+                    node.getProperty(AuthoringPublicationConstant.LIVE_REVISION_PROP).getString() : null;
+            String versionName = node.getVersionHistory().getSession().getNodeByUUID(nodeVersionUUID).getName();
+            if (!versionName.isEmpty()) {
+              node.restore(versionName, true);
+              if (!node.isCheckedOut()) {
+                node.checkout();
+              }
+              publicationService.changeState(node, "published", new HashMap<>());
```
azayati:
```suggestion
              publicationService.changeState(node, PublicationDefaultStates.PUBLISHED, new HashMap<>());
```

azayati (changes requested):
See review

azayati (approved):
LGTM

PR #148, opened 2021-06-15
Task-45328: Implement Publish news drawer from the new editor 

hassenelkamel-exo:
Prior to this change, from news editor when creating a news, we can post directly the news. After this change, posting a news should be done using a publish drawer

### DIFF
--- a/webapp/src/main/webapp/news-activity-composer-app/components/ExoNewsActivityComposer.vue
+++ b/webapp/src/main/webapp/news-activity-composer-app/components/ExoNewsActivityComposer.vue
@@ -1,5 +1,5 @@
 <template>
-  <div>
+  <div class="VuetifyApp">
     <div v-show="loading">
       <div class="VuetifyApp loadingComposer">
         <v-progress-circular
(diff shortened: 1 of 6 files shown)


webapp/src/main/webapp/news-activity-composer-app/components/ExoNewsActivityComposer.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <div>
+  <div class="VuetifyApp">
     <div v-show="loading">
       <div class="VuetifyApp loadingComposer">
```
azayati: do you need really the class VuetifyApp in each block ?
If you keep it only on the parent div, will it be sufficient ?

webapp/src/main/webapp/news-activity-composer-app/components/ExoNewsActivityComposer.vue
```diff
@@ -44,31 +45,35 @@
...
-              :disabled="news.archived ? true: updateDisabled"
-              :class="[news.archived ? 'unauthorizedPin' : '']"
-              class="btn"
-              @click.prevent="updateAndPostNews">
-              {{ $t("news.edit.update.post") }}
-            </button>
-            <button class="btn" @click="goBack">
-              {{ $t("news.composer.btn.cancel") }}
-            </button>
+          <div v-show="editMode" class="VuetifyApp">
```
azayati: do you need really the class VuetifyApp in each block ?
If you keep it only on the parent div, will it be sufficient ?

azayati (changes requested):
See review

azayati (approved):
LGTM after QA validation

PR #155, opened 2021-06-18
Task-45975: The latest news CLV disappears after deleting a pinned news

AzmiTouil:
Before this fix, when we delete a pinned news, its associated node is deleted without removing the related symlink generated after pin from the application data.
In order to be on a coherent jcr state and to endure the display of latest news portlet, we need to remove all symlinks of the removed news node.


azayati (approved):
LGTM after QA validation

PR #158, opened 2021-06-22
Task-45984: Fix forEach not a function error on js HTMLCollection when display news details

azayati:
Prior to this change, in some cases forEach applied on HTMLCollection returned by getElementsByTagName function make a not function error. So, we replace that by a classic for loop in order to avoid such errors.


AzmiTouil (approved):
LGTM

PR #161, opened 2021-06-23
Task-46089: Fix display alert when deleting news from news details

AzmiTouil:
This PR will fix the regression in order to display again alert when deleting news in news detail.


azayati (approved):
LGTM after QA validation

PR #162, opened 2021-06-23
Task-45534: Fix pinned article detail back button redirection for non news space members 

AzmiTouil:
Prior to this change, when we click on back button from a pinned article detail as non news space member and non super spaces or user admin, we are redirected to non-acces-space page. This PR will update the redirection to the snapshot page in that case.

### DIFF
--- a/webapp/src/main/webapp/news-details/components/ExoNewsDetails.vue
+++ b/webapp/src/main/webapp/news-details/components/ExoNewsDetails.vue
@@ -199,7 +199,7 @@ export default {
       return this.news && (this.news.profileAvatarURL || this.news.authorAvatarUrl);
     },
     backURL() {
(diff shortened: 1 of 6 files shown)


webapp/src/main/webapp/news-details/components/ExoNewsDetails.vue
```diff
@@ -198,6 +198,9 @@ export default {
     authorAvatarURL() {
       return this.news && (this.news.profileAvatarURL || this.news.authorAvatarUrl);
     },
+    backURL() {
+      return this.news && this.news.url.includes('/news/detail?content-id=') ? `${eXo.env.portal.context}/${eXo.env.portal.portalName}` : this.news.spaceUrl;
```
azayati: I prefer to check correctly if the user is a member of the news space. We can not base on urls check because it can change and in that case we should return here again to follow.

azayati (changes requested):
See review

azayati (approved):
LGTM after QA validation

PR #165, opened 2021-06-25
Task-46183: Optimize news related notifications

AzmiTouil:
Prior to this change, we we share news, like or comment news activity, notifications are sent. This PR will keep only post and mention news notifications.


azayati (approved):
LGTM after QA validation

PR #181, opened 2021-07-07
Task-46550: Sort the news search result in the news app by the modifcation date

AzmiTouil:
Prior to this change, the news search result in the news app is sorted by **jcr:score**. This PR we will allow sorting by the modification date.


azayati (approved):
LGTM

PR #186, opened 2021-07-09
Task-46742: Fix displaying posted articles issue

azayati:
Prior to this change and after merging [PR ](https://github.com/exoplatform/news/pull/180), anyone can see all news from posted article filter of news app even if he is not allowed. The jcr session used in wrong, with this PR, anyone see only allowed posted articles.


hassenelkamel-exo (approved):
LGTM

PR #190, opened 2021-07-13
Task-46050: Implement news analytics board

AzmiTouil:
Prior to this change, we are not able to calculate metrics for news operations. This PR will implement news analytics board


services/src/main/java/org/exoplatform/news/listener/NewsGamificationIntegrationListener.java
```diff
@@ -57,8 +51,8 @@ public void onEvent(Event<String, String> event) throws Exception {
         Map<String, String> gamificationMap = new HashMap<>();
         gamificationMap.put("ruleTitle", ruleTitle);
         gamificationMap.put("object", news.getUrl());
-        gamificationMap.put("senderId", earnerId); // matches the gamification's earner id
-        gamificationMap.put("receiverId", earnerId);
+        gamificationMap.put("senderId", news.getAuthor()); // matches the gamification's earner id
+        gamificationMap.put("receiverId", news.getAuthor());
```
ahamdi, line 55: I think we need just the senderId , no need to store the receiverId right ?

services/src/test/java/org/exoplatform/news/rest/NewsRestResourcesV1Test.java
```diff
@@ -1954,4 +1955,62 @@ public void shouldGetStagedNewsWhenCurrentUserIsAuthor() throws Exception {
     assertEquals("john", newsList.get(0).getAuthor());
   }
 
+  @Test
+  public void getNewsIllustrationTest() {
```
ahamdi, line 2062: UT are not related to the added code. I understand that you want to keep the same test coverage. That is fine

services/src/main/java/org/exoplatform/news/listener/NewsGamificationIntegrationListener.java
AzmiTouil: The gamification object map must contain both the senderId and receiverId params.

ahamdi (dismissed):

ahamdi (dismissed):

azayati (approved):
LGTM

PR #199, opened 2021-07-23

hassenelkamel-exo:
- When you use` sessionProviderService.getSessionProvider(null)` : it throws null pointer exception, so we use
 `sessionProviderService.getSystemSessionProvider(null) ` to avoid this problem.


ahamdi (changes requested):
We should not use a System session for a user operation.

hassenelkamel-exo:
> We should not use a System session for a user operation.
IMO, it isn't a user operation, this used function will be reached only when a notification sent.
how to solve the problem of null pointer exception, i think we don't have another solution in this case

PR #204, opened 2021-07-26
task-47214: hide article attachments in news details when detected by DLP

AzmiTouil:
When an attached document to an article is detected by DLPs the document should not be available from the article's details


ahamdi:
@AzmiTouil Please upadte the Test case to cover the quaratined files case

ahamdi (approved):

ahamdi (changes requested):
This should be proposed for feature/maintenance and nor develop

PR #206, opened 2021-07-27
Task-46673: Unify thumbnails of news App and news content list

hassenelkamel-exo:
**Problem:**  In news App and news content list we display the old thumbnail.
**Solution:**  We unified news app and news content list default thumbnail with app center one.


azayati (approved):
Please update description with problem and solution
LGTM after QA validation

PR #210, opened 2021-07-29

hassenelkamel-exo:
**-Problem:**  When we use `sessionProviderService.getSessionProvider(null)` : it throws null pointer exception, when `NewsPublicationListener` launch.
**-Solution:**  We create 2 methods for news creation: the first one that use `getSessionProvider` , the second calls by `NewsPublicationListener` and use system session provider

### DIFF
--- a/services/src/main/java/org/exoplatform/news/NewsServiceImpl.java
+++ b/services/src/main/java/org/exoplatform/news/NewsServiceImpl.java
@@ -196,12 +196,40 @@ public class NewsServiceImpl implements NewsService {
    * @throws Exception when error
    */
   public News createNews(News news) throws Exception {
+    SessionProvider sessionProvider = sessionProviderService.getSessionProvider(null);
+    Session session = sessionProvider.getSession(
+                                                 repositoryService.getCurrentRepository()
+                                                                  .getConfiguration()
+                                                                  .getDefaultWorkspaceName(),
+                                                 repositoryService.getCurrentRepository());
     if (StringUtils.isEmpty(news.getId())) {
       news = createNewsDraft(news);
     } else {
-      postNewsActivity(news);
+      postNewsActivity(news, session);
       updateNews(news);
-      sendNotification(news, NotificationConstants.NOTIFICATION_CONTEXT.POST_NEWS);
+      sendNotification(news, NotificationConstants.NOTIFICATION_CONTEXT.POST_NEWS, session);
+    }
+
+    if (news.isPinned()) {
+      pinNews(news.getId());
+    }
+    NewsUtils.broadcastEvent(NewsUtils.POST_NEWS, news.getId(), news.getAuthor());
+    return news;
+  }
+
+  public News createNews(News news, Session session) throws Exception {
+    if (StringUtils.isEmpty(news.getId())) {
+      news = createNewsDraft(news);
+    } else {
+      try {
+        postNewsActivity(news, session);
+        updateNews(news);
+        sendNotification(news, NotificationConstants.NOTIFICATION_CONTEXT.POST_NEWS, session);
+      } finally {
+        if (session != null) {
+          session.logout();
+        }
+      }
     }
 
     if (news.isPinned()) {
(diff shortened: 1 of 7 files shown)


services/src/main/java/org/exoplatform/news/NewsServiceImpl.java
```diff
@@ -196,12 +196,40 @@ public NewsServiceImpl(RepositoryService repositoryService,
...
+      news = createNewsDraft(news);
+    } else {
+      try {
+        postNewsActivity(news, session);
+        updateNews(news);
+        sendNotification(news, NotificationConstants.NOTIFICATION_CONTEXT.POST_NEWS, session);
+      } finally {
+        if (session != null) {
+          session.logout();
+        }
```
ahamdi: Session should be closed in the same function where it was created.

ahamdi (changes requested):

services/src/main/java/org/exoplatform/news/NewsServiceImpl.java
hassenelkamel-exo: I forgot it,no need i will remove it ,

services/src/main/java/org/exoplatform/news/NewsServiceImpl.java
```diff
@@ -196,12 +196,34 @@ public NewsServiceImpl(RepositoryService repositoryService,
...
       updateNews(news);
-      sendNotification(news, NotificationConstants.NOTIFICATION_CONTEXT.POST_NEWS);
+      sendNotification(news, NotificationConstants.NOTIFICATION_CONTEXT.POST_NEWS, session);
+    }
+
+    if (news.isPinned()) {
+      pinNews(news.getId());
+    }
+    NewsUtils.broadcastEvent(NewsUtils.POST_NEWS, news.getId(), news.getAuthor());
+    return news;
```
ahamdi, line 223: We need to close the session provider after ending with its usage.

services/src/main/java/org/exoplatform/news/listener/NewsPublicationListener.java
```diff
@@ -32,19 +38,32 @@
...
     if (AuthoringPublicationConstant.POST_CHANGE_STATE_EVENT.equals(event.getEventName())) {
       Node targetNode = event.getData();
       if (targetNode.isNodeType("exo:news") && targetNode.getProperty(StageAndVersionPublicationConstant.CURRENT_STATE).getString().equals(PublicationDefaultStates.PUBLISHED)) {
         News news = newsService.convertNodeToNews(targetNode, false);
         if (StringUtils.isEmpty(news.getActivities())) {
-          newsService.createNews(news);
+          newsService.createNews(news, session);
         }
       }
     }
```
ahamdi, line 73: We need to close the session provider here

ahamdi (changes requested):

services/src/main/java/org/exoplatform/news/listener/NewsPublicationListener.java
hassenelkamel-exo: done

services/src/main/java/org/exoplatform/news/NewsServiceImpl.java
hassenelkamel-exo: done

services/src/main/java/org/exoplatform/news/NewsServiceImpl.java
```diff
@@ -196,12 +196,46 @@ public NewsServiceImpl(RepositoryService repositoryService,
...
-      updateNews(news);
-      sendNotification(news, NotificationConstants.NOTIFICATION_CONTEXT.POST_NEWS);
+      try {
+        postNewsActivity(news, session);
+        updateNews(news);
+        sendNotification(news, NotificationConstants.NOTIFICATION_CONTEXT.POST_NEWS, session);
+      } finally {
+        if (session != null) {
+          session.logout();
+        }
```
ahamdi: Session is created outside here, it may still be used. we should not close it here

hassenelkamel-exo: I think that is no problem to close here, in the case to pin it, in pinNew, we have opened another session there,

hassenelkamel-exo: session closed in NewsPublicationListener

services/src/main/java/org/exoplatform/news/listener/NewsPublicationListener.java
```diff
@@ -32,19 +38,37 @@
...
       if (targetNode.isNodeType("exo:news") && targetNode.getProperty(StageAndVersionPublicationConstant.CURRENT_STATE).getString().equals(PublicationDefaultStates.PUBLISHED)) {
         News news = newsService.convertNodeToNews(targetNode, false);
         if (StringUtils.isEmpty(news.getActivities())) {
-          newsService.createNews(news);
+          try {
+            newsService.createNews(news, session);
+          } finally {
+            if (session != null) {
+              session.logout();
+            }
```
ahamdi: Session  won't be closed unless the above checks are valid. We need to extract it outside all if block

ahamdi (approved):

PR #219, opened 2021-08-05
Task-47085: Fix session timeout when creating news

hassenelkamel-exo:
Prior to this change, the jcr session is always closed in `NewsPublicationListener` and since this listener is launched when we create draft news, we have a session timeout. With this PR, we ensure that the listener is not launched for draft news creation.


ahamdi (dismissed):

azayati (approved):
LGTM

PR #224, opened 2021-08-09
MAINT-38206: Fix latest news header update in CLV portlet settings

hakermi:
Fix latest news header update in CLV portlet settings


azayati (approved):
LGTM

azayati:
@Jihed525 @hakermi it should be better to keep the original description of the PR instead of cherry-picked from cc @ahamdi @hbenali

PR #243, opened 2021-08-18

hassenelkamel-exo:
**-Problem:** When we use sessionProviderService.getSessionProvider(null) : it throws null pointer exception, when NewsPublicationListener launch.
**-Solution:** We create 2 methods for news creation: the first one that use getSessionProvider , the second calls by NewsPublicationListener and use system session provider


azayati (approved):
LGTM

azayati (approved):
LGTM

PR #247, opened 2021-08-18
Task-47867_47868: Wrong Daily/Weekly notification for posted and published news

hassenelkamel-exo:
**Problem:**  We don't get the content of all notifications because we use `map.entrySet()` which all time return size 1 for post and publish notification type.
**Solution:**  We must get the content from the whole notification list.


azayati (approved):
LGTM

AzmiTouil (approved):
LGTM

PR #256, opened 2021-08-25
Task-48190 : The switch to an article button shouldn't be displayed when exceed 1300 chars in stream main page

ahmadmido:
Issue: When adding a post with exceeding 1300 chars, the switch to an article button is displayed on mainstream and space stream.
Solution: I have added a check to verify if is mainstream or space stream, and I display the switch to an article button only on space stream.


rdenarie (dismissed):
Ok for me
On remark aboud commit message : can you try to have one line for the issue and one for the solution, it is more reable.
In addition can you try to use imperative form (this change modifies ...) instead of using I

rdenarie (approved):

PR #260, opened 2021-08-31
Task-48372: Fix wrong redirection from News details page to space page stream.

ahmadmido:
Issue: before this fix, when we click on the return button from the news article, the user is redirected to the Snapshot page.
Solution: the fix will ensure that the right variable is checked on the news object( spaceMember)


azayati (approved):
LGTM

PR #261, opened 2021-09-01
Task-48190 : The switch to an article button shouldn't be displayed when exceed 1300 chars in stream main page.

ahmadmido:
Issue: When adding a post with exceeding 1300 chars, the switch to an article button is displayed on mainstream and space stream.
Solution: Verify if the property eXo.env.portal.spaceId isn't empty (empty in case of main stream), and check if switchToArticleActivityComposerPlugin.enabled on space stream to display the switch to an article button.

### DIFF
--- a/webapp/src/main/webapp/news-extensions/extensions.js
+++ b/webapp/src/main/webapp/news-extensions/extensions.js
@@ -95,7 +95,7 @@ export function initExtensions() {
   if (eXo.env.portal.spaceId) {
     canUserCreateNews(eXo.env.portal.spaceId)
       .then(canCreateNews => {
-        switchToArticleActivityComposerPlugin.enabled = newsActivityComposerPlugin.enabled = eXo.env.portal.spaceId && canCreateNews;
+        switchToArticleActivityComposerPlugin.enabled = newsActivityComposerPlugin.enabled = (eXo.env.portal.spaceId && eXo.env.portal.spaceId !== '') && canCreateNews;
         extensionRegistry.registerExtension('ActivityComposer', 'activity-composer-action', newsActivityComposerPlugin);
         document.dispatchEvent(new CustomEvent('activity-composer-extension-updated'));
       });


webapp/src/main/webapp/news-extensions/extensions.js
```diff
@@ -95,7 +95,7 @@ export function initExtensions() {
   if (eXo.env.portal.spaceId) {
     canUserCreateNews(eXo.env.portal.spaceId)
       .then(canCreateNews => {
-        switchToArticleActivityComposerPlugin.enabled = newsActivityComposerPlugin.enabled = eXo.env.portal.spaceId && canCreateNews;
+        switchToArticleActivityComposerPlugin.enabled = newsActivityComposerPlugin.enabled = (eXo.env.portal.spaceId && eXo.env.portal.spaceId !== '') && canCreateNews;
```
ahamdi, line 98: I am not sure that this check is useful
`eXo.env.portal.spaceId !== ''`
check https://developer.mozilla.org/en-US/docs/Glossary/Falsy

ahamdi (changes requested):

webapp/src/main/webapp/news-extensions/extensions.js
ahmadmido: > I am not sure that this check is useful
> `eXo.env.portal.spaceId !== ''`
> check https://developer.mozilla.org/en-US/docs/Glossary/Falsy

I think it's necessary to check if `eXo.env.portal.spaceId !== ''`, The main stream has an empty value on `eXo.env.portal.spaceId  so we should do this check.

ahamdi:
```suggestion
        switchToArticleActivityComposerPlugin.enabled = newsActivityComposerPlugin.enabled = eXo.env.portal.spaceId && canCreateNews;
```

PR #269, opened 2021-09-06
Task-46669: Fix image position in news body

hassenelkamel-exo:
**Problem**:  When we change device the news body image position is not responsive
**Solution**: Use vw and rem instead of percent and px to make image responsive when changing device

### DIFF
--- a/webapp/src/main/webapp/skin/less/news.less
+++ b/webapp/src/main/webapp/skin/less/news.less
@@ -41,6 +41,7 @@
       border: none !important;
       float: right ~'; /** orientation=lt */ ';
       float: left ~'; /** orientation=rt */ ';
+      margin-left: 15px ~'; /** orientation=rt */ ';
       i {
         font-size: 18px;
       }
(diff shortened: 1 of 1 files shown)


webapp/src/main/webapp/skin/less/news.less
```diff
@@ -175,13 +175,13 @@
     }
 
     img.pull-right {
-      margin-left: 15px ~'; /** orientation=lt */ ';
-      margin-right: 15px ~'; /** orientation=rt */ ';
+      margin-left: 1.5rem ~'; /** orientation=lt */ ';
```
azayati, line 179: RTL to be tested @hassenelkamel-exo

azayati (changes requested):
See review

webapp/src/main/webapp/skin/less/news.less
hassenelkamel-exo: yes, well tested

azayati (approved):
LGTM after QA validation

PR #273, opened 2021-09-07
Task-42873: Change the published news detail page when accessible by non space members

hassenelkamel-exo:
**Problem:**  When displaying published article for non space member, the tabs content  is `News details`.
**Solution:**  When getting the news to be displayed, we set the right document title `News: title of article`

### DIFF
--- a/webapp/src/main/webapp/news-details/components/ExoNewsDetails.vue
+++ b/webapp/src/main/webapp/news-details/components/ExoNewsDetails.vue
@@ -289,6 +289,7 @@ export default {
           return this.$nextTick();
         })
         .finally(() => {
+          document.title = this.$t('news.window.title', {0: this.news.title});
           this.$root.$emit('application-loaded');
         });
     } else {


azayati (approved):
LGTM after QA validation

webapp/src/main/webapp/news-details/components/ExoNewsDetails.vue
```diff
@@ -289,6 +289,7 @@ export default {
           return this.$nextTick();
         })
         .finally(() => {
+          document.title = this.$t('news.window.title', {0: this.news.title});
```
azayati, line 292: Should you ensure that you display only the 100 first chars of the news title ?

azayati (changes requested):
See review

webapp/src/main/webapp/news-details/components/ExoNewsDetails.vue
hassenelkamel-exo: It display the same like the activity

azayati (approved):
LGTM

PR #282, opened 2021-09-17
Task-47159: Add the possibility to publish an article when editing a schedule to be posted immediately or later.

hassenelkamel-exo:
Prior to this change, it is not possible to choose to publish an article when editing a schedule to be posted immediately or later.
After this change, we add the possibility to publish an article when editing a schedule. After checking publish news checkbox, the article will be published (unpublished) when posted immediately or scheduled for being posted later.


azayati (approved):
LGTM

PR #283, opened 2021-09-20
Task-49049: Fix the content of popover when hovering from news details

hassenelkamel-exo:
Prior to this change,  in news details, when hovering the author, it displays a popover with information about the last modifier. The solution is to get the url of news author instead of news updater.


azayati (approved):
LGTM after QA validation

PR #288, opened 2021-09-24
Task-49368: Fix regression on edit/delete/resume articles from news app 

hassenelkamel-exo:
Prior to this change, after refactoring done in mobile view `ExoNewsDetailsActionMenu` isn't the child component of NewsApp, the deletion and edition action doesn't work. So to make this action possible, we have created a child component `ExoNewsDetailsActionMenuApp` allowing edition and deletion.


azayati (approved):
LGTM

PR #289, opened 2021-09-24
Task-49408: Fix exception when liking or commenting a non news or shared news activity types

azayati:
Prior to this change, when we like or comment any activity type except news activity or shared news activity types we have this exception:
ERROR | Error broadcast comment news event [o.e.news.listener.NewsActivityListener<http-nio-0.0.0.0-8080-exec-6>] org.exoplatform.commons.exception.ObjectNotFoundException: Activity with id 2 isn't of type news nor a shared new
since likeActivity and saveComment methods of NewsActivityListener are invoked on any type of activity and getNewsByActivityId generate the exception when the activity is not of news or shared news type.
After this change, we ensure that the treatment of these methods is done only for news activity or shared news activity types.


ahmadmido (approved):

hassenelkamel-exo (approved):
LGTM

PR #290, opened 2021-09-27
Task-49461: Fix publish/unpublish labels and actions regressions in news details desktop and mobile

hassenelkamel-exo:
Prior to change, the publish news confirmation popup labels, publish icon tooltip label are not correct and the publish/unpublish actions on desktop and mobile doesn't work. After this change, the problems are fixed.


azayati (dismissed):
LGTM until QA validation

azayati (approved):
LGTM

PR #291, opened 2021-09-28
Backport Task-46549_49049: Implement mobile view for news detail and Fix the content of popover when hovering from news details

azayati:
Prior to change, news details was created in one component and when hovering the author, it displays a popover with information about the last modifier. After this change, in order to make code more maintainable and to implement mobile view, we refactor the code by splitting into components specific to mobile view, others specific to desktop view and other common components to mobile and desktop views. We have also fixed the popover problem.


hassenelkamel-exo (approved):
LGTM

PR #294, opened 2021-09-30
Task-49273: Display drawer attachment when clicking on attach ckeditor button with attached files number and files details in news details mobile view

hassenelkamel-exo:
Prior to this change, we use attach button in the bottom of ckeditor. With this enhancement, we add an attach button to `ckeditor` toolbar which will open attachment drawer, the number of attached files will be shown in the bottom of the attach button and the attached files details will be displayed.


azayati (changes requested):
PR title and description modified, check them @hassenelkamel-exo

azayati (approved):
LGTM

PR #299, opened 2021-10-04
Task-47385: Implement mobile view for news editor when editing news

hassenelkamel-exo:
Prior to this change, in mobile view the news composer when editing news is not well displayed. In order to enhance view mobile, we have removed update and post button and restyled the UI.


azayati (approved):
LGTM after QA validation

PR #304, opened 2021-10-07
Task-47951: Update News Editor, CKeditor toolbar and placeholder colors

ahmadmido:
Prior to this change, News Editor, CKeditor toolbar and placeholder colors are too ugly and not comfortable on the eyes. After this change these colors are updated.


azayati (approved):
LGTM after QA validation

PR #312, opened 2021-10-20
Task-50136: Fix shared draft edition by a space redactor

ahmadmido:
Prior to this fix, a space redactor can't edit a shared draft, all fields are empty. This fix will ensure that a space redactor can edit a shared draft.


services/src/main/java/org/exoplatform/news/NewsServiceImpl.java
```diff
@@ -1198,6 +1198,9 @@ public boolean canEditNews(News news, String authenticatedUser) {
     if (spaceService.isManager(currentSpace, authenticatedUser)) {
       return true;
     }
+    if (spaceService.isRedactor(currentSpace, authenticatedUser) && news.isDraftVisible()) {
```
azayati: I think that this condition should be verified for a posted article draft which was draft visible before posting, normally the redactors in that case should be able to edit the article

azayati (changes requested):
See review

azayati (dismissed):
LGGTM after QA validation

azayati (approved):
LGTM

PR #313, opened 2021-10-20
Task-49720:  Add the possibility to Post or not Post on the stream

hassenelkamel-exo:
Prior to this change, when we post an article, a news activity is generated automatically. After this change,, we add the possibility to choose the generated news activity is posted or not posted on the stream. When we choose to not post, the generated activity is hidden.

### DIFF
--- a/services/src/main/java/org/exoplatform/news/NewsServiceImpl.java
+++ b/services/src/main/java/org/exoplatform/news/NewsServiceImpl.java
@@ -841,7 +841,6 @@ public class NewsServiceImpl implements NewsService {
     activity.isHidden(news.isHiddenActivity());
     Map<String, String> templateParams = new HashMap<>();
     templateParams.put("newsId", news.getId());
-    templateParams.put("isHiddenActivity", String.valueOf(news.isHiddenActivity()));
     activity.setTemplateParams(templateParams);
 
     activityManager.saveActivityNoReturn(spaceIdentity, activity);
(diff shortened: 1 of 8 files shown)


services/src/main/java/org/exoplatform/news/NewsServiceImpl.java
```diff
@@ -836,8 +838,10 @@ void postNewsActivity(News news, Session session) throws Exception {
     activity.setBody("");
     activity.setType("news");
     activity.setUserId(poster.getId());
+    activity.isHidden(news.isHiddenActivity());
     Map<String, String> templateParams = new HashMap<>();
     templateParams.put("newsId", news.getId());
+    templateParams.put("isHiddenActivity", String.valueOf(news.isHiddenActivity()));
```
azayati: we don't need that IMO

azayati (changes requested):
See review
Keep only modifications related to post/don't post on the stream

services/src/main/java/org/exoplatform/news/NewsServiceImpl.java
hassenelkamel-exo: yes i will remove it

webapp/src/main/webapp/schedule-news-drawer/components/ExoScheduleNewsDrawer.vue
```diff
@@ -288,7 +289,7 @@ export default {
     disabled() {
       const postDate = new Date(this.postDate);
       const scheduleDate = new Date(this.schedulePostDate);
-      return (this.postArticleMode === 'immediate' ? false : this.postArticleMode === 'later' && postDate.getTime() === scheduleDate.getTime()) && this.selected === this.publish;
+      return (this.postArticleMode === 'immediate' ? false : this.postArticleMode === 'later' && postDate.getTime() === scheduleDate.getTime()) && this.selected === this.publish || this.stepper<3;
```
azayati:
```suggestion
      return (this.postArticleMode === 'immediate' ? false : this.postArticleMode === 'later' && postDate.getTime() === scheduleDate.getTime()) && this.selected === this.publish || this.stepper < 3;
```

azayati (changes requested):
See review

azayati (approved):
LGTM after QA validation

PR #314, opened 2021-10-20
Task-50365: Fix draft visibility selector label display from news details

ahmadmido:
Prior to this change, the draft visibility selector label is not well displayed in one line when the label is too long after translation. After this change, the display is well fixed in that case


azayati (approved):
LGTM after QA validation

PR #317, opened 2021-10-22
Task-50396: Use specific draft deletion confirmation messages

azayati:
Prior to this change, when deleting a draft from news app, we use the same confirmation messages as for article deletion. After this change, we will use specific confirmation messages for draft deletion.


hassenelkamel-exo (approved):
LGTM

ahmadmido (approved):

PR #318, opened 2021-10-25
Task-50098: The article is not raised to the top of the space when it is edited by a non owner with the option update and post.

ahmadmido:
Prior to this change, when editing a news by someone else not the owner with update and post, the related activity is not raised to the top of the activity stream. After this fix, we will ensure that when editing a news with update and post, the activity is updated and raised to the top of the activity stream.

### DIFF
--- a/services/src/main/java/org/exoplatform/news/NewsServiceImpl.java
+++ b/services/src/main/java/org/exoplatform/news/NewsServiceImpl.java
@@ -1191,6 +1191,15 @@ public class NewsServiceImpl implements NewsService {
     return authenticatedSecurityIdentity.isMemberOf(PLATFORM_WEB_CONTRIBUTORS_GROUP, PUBLISHER_MEMBERSHIP_NAME);
   }
 
+  @Override
+  public void updateNewsActivity(News news, boolean post) {
+    ExoSocialActivity activity = activityManager.getActivity(news.getActivityId());
+    if(post) {
+      activity.setUpdated(System.currentTimeMillis());
+    }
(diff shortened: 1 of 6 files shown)


services/src/main/java/org/exoplatform/news/rest/NewsRestResourcesV1.java
```diff
@@ -478,6 +481,13 @@ public Response updateNews(@Context HttpServletRequest request,
 
       news = newsService.updateNews(news);
 
+      if(post) {
+        ActivityManager activityManager = CommonsUtils.getService(ActivityManager.class);
```
azayati: add that in the constructor

azayati (changes requested):
See review

services/src/main/java/org/exoplatform/news/NewsServiceImpl.java
```diff
@@ -1191,6 +1191,15 @@ public boolean canEditNews(News news, String authenticatedUser) {
     return authenticatedSecurityIdentity.isMemberOf(PLATFORM_WEB_CONTRIBUTORS_GROUP, PUBLISHER_MEMBERSHIP_NAME);
   }
 
+  @Override
+  public void updateActivityNews(News news, boolean post) {
```
azayati:
```suggestion
  public void updateNewsActivity(News news, boolean post) {
```

azayati (changes requested):
see review

azayati (approved):
LGTM

PR #326, opened 2021-10-28
Task-50470: Don't display posted article edited draft warning alert message for current user

ahmadmido:
Prior to this fix, a warning alert message is displayed when editing the posted article and the current user is its last modifier.
This fix will ensure that the warning alert message is displayed only when the last modifier is not the current user.


azayati (approved):
LGTM after QA validation

PR #327, opened 2021-10-28
Task-50376_50599: News publisher and externals shouldn't receive published news notification

ahmadmido:
Prior to this change, news publisher and externals receive published news notification. After this change, only users except news publisher should receive published news notifications.


services/src/main/java/org/exoplatform/news/notification/plugin/PublishNewsNotificationPlugin.java
```diff
@@ -48,15 +65,23 @@ protected NotificationInfo makeNotification(NotificationContext ctx) {
...
+    try {
+      receivers = getReceivers(contentSpaceId, currentUserName);
+    } catch (Exception e) {
+      LOG.error("An error occured when trying to have the list of receivers " + e.getMessage(), e);
+    }
+
     return NotificationInfo.instance()
                            .setFrom(currentUserName)
-                           .setSendAll(true)
+                           .to(receivers)
```
azayati: the audience of publishing is every user except externals and publisher itself

azayati (changes requested):
see review

services/src/main/java/org/exoplatform/news/notification/plugin/PublishNewsNotificationPlugin.java
azayati: I think that it would be better to :

- keep using .setSendAll(true)
- implement exclude(List users) method based on removeOnSendToDaily and removeOnSendToWeekly which will be used to exclude externals and current user

services/src/main/java/org/exoplatform/news/notification/plugin/PublishNewsNotificationPlugin.java
```diff
@@ -70,4 +90,21 @@ protected NotificationInfo makeNotification(NotificationContext ctx) {
                            .end();
 
   }
+
+  private List<String> getReceivers(String currentUserName) throws Exception {
+    Group group = getOrganizationService().getGroupHandler().findGroupById(PLATFORM_USERS);
+    Set<String> receivers = new HashSet<String>();
+    ListAccess<Membership> membershipsListAccess = organizationService.getMembershipHandler()
```
azayati: You can use UserHandler.findUsersByGroupId instead

azayati (changes requested):
See review

services/src/main/java/org/exoplatform/news/notification/plugin/PublishNewsNotificationPlugin.java
```diff
@@ -70,4 +90,15 @@ protected NotificationInfo makeNotification(NotificationContext ctx) {
                            .end();
 
   }
+
+  private List<String> getReceivers(String currentUserName) throws Exception {
```
azayati: we should rework that from notification API of commons project, please follow this [poc](https://github.com/Meeds-io/commons/compare/notification-for-internal-users-poc)

azayati (changes requested):
see review

azayati (approved):
LGTM

PR #328, opened 2021-10-29
Task-50717: Fix exceptions when creating news

hassenelkamel-exo:
Prior to this change, before creating news we have a null pointer exception due to a null value of activity. In addition, after posting news, we have some exceptions related to gamification and analytics due to problem of the wrong type of variable used in `broadcastEvent` and the use the same key for gamification rule and analytics rule. Also, when scheduling a news, we get a null value for the username when broadcasting the analytics event so we will send the author of news.


azayati (dismissed):
LGTM after QA validation

azayati (dismissed):
LGTM after QA validation

azayati (dismissed):
LGTM

azayati (approved):
LGTM

PR #330, opened 2021-11-01
Task-50744: Remove the news publishing options for admins

hassenelkamel-exo:
Prior to this change, administrators can publish articles and use the new scheduling news stepper when they are simple space members. After this change, the publishing and scheduling options will not be available for administrators.


services/src/main/java/org/exoplatform/news/NewsServiceImpl.java
```diff
@@ -95,8 +95,6 @@
 
   private final static String      PLATFORM_WEB_CONTRIBUTORS_GROUP  = "/platform/web-contributors";
```
azayati, line 96: you should also modify canReactOnSpace from social in order to not allow admins to react on redactional space

azayati (changes requested):
see review

azayati (approved):
LGTM after QA validation

PR #332, opened 2021-11-02
Task-49697: Fix button id when fixing colorButton plugin regression

hassenelkamel-exo:
Prior to this change, we can not apply a color for a text from ckeditor. This problem is due to the absence of colorbutton plugin.


azayati (dismissed):
LGTM after QA validation

azayati (approved):
LGTM

PR #336, opened 2021-11-08
Task-50098: The article is not raised to the top of the space when it is edited by a non owner with the option update and post. (#318)

ahmadmido:
Prior to this change, when editing news by someone else not the owner with updates and posts, the related activity is not raised to the top of the activity stream. After this fix, we will ensure that when editing news with updates and posts, the activity is updated and raised to the top of the activity stream.

(cherry picked from commit 8cdb15458c25fb8fde00e5d64d388c86eeb7f075)


azayati (approved):
LGTM

PR #339, opened 2021-11-09
Task-51114: Fix error when editing a scheduled news by a redactor different from the news owner

hassenelkamel-exo:
Prior to this change, when editing a scheduled news by a redactor different from the news owner. To solve this problem we will modify `canEditNews` method in order to allow redactor to update scheduled news.


azayati (approved):
LGTM after QA validation

PR #346, opened 2021-11-16
Task-49412: Ensure to refresh the news application after deleting an article.

ahmadmido:
Prior to this change, when we delete an article from news application, the news list isn't well refreshed. The problem occurs with a big number of articles since the delete isn't done yet from the database. With this fix, we have increased the redirectionTime to make sure the news list refresh is done after article deletion.


azayati (approved):
LGTM after QA validation

PR #356, opened 2021-11-24
Task-51649: Don't post news activity when editing an unposted article

hassenelkamel-exo:
Prior to this change when updating an unposted article, an activity is created. After this change we add the status postedActivity to keep the status of activity.


azayati (dismissed):
LGTM after QA validation

azayati (approved):
LGTM

PR #357, opened 2021-11-25
TASK-51421: Bookmark/unbookmark a news

aelbenney:


nidhalnidhal (approved):

azayati (changes requested):
We have a big refactoring in progress, please don't merge anything on develop before merging the refactoring

services/src/main/java/org/exoplatform/news/rest/NewsRestResourcesV1.java
```diff
@@ -906,4 +919,43 @@ private boolean canScheduleNews(String authenticatedUser, Space space) {
     return spaceService.isManager(space, authenticatedUser) || spaceService.isRedactor(space, authenticatedUser) || currentIdentity.isMemberOf(PLATFORM_WEB_CONTRIBUTORS_GROUP, PUBLISHER_MEMBERSHIP_NAME);
   }
 
+
+  public static Map<String, List<MetadataItemEntity>> retrieveMetadataItems(News news,
```
azayati, line 922: We should not call directly Metadata service from rest layer neither from service layer.
Rest layer should call service layer and service layer should call favorite service

azayati (changes requested):
See review

PR #358, opened 2021-11-29
Task-50931: Implement getReferencedTargets API

ahmadmido:
In this PR we will Implement getReferencedTargets API in order to retrieve referenced news targets in news list portlets

### DIFF
--- a/services/src/test/java/org/exoplatform/news/NewsTargetingImplTest.java
+++ b/services/src/test/java/org/exoplatform/news/NewsTargetingImplTest.java
@@ -64,6 +64,28 @@ public class NewsTargetingImplTest {
     assertEquals("latest news", newsTargetingEntities.get(1).getLabel());
   }
 
+  @Test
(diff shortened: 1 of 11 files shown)


services/src/test/java/org/exoplatform/news/NewsTargetingImplTest.java
```diff
@@ -64,4 +64,29 @@ public void shouldReturnTargets() {
     assertEquals("latest news", newsTargetingEntities.get(1).getLabel());
   }
 
+
+  @Test
+  public void shouldReturnReferencedTargets() {
+    // Given
+    NewsTargetingServiceImpl newsTargetingService = new NewsTargetingServiceImpl(metadataService);
```
azayati, line 93: can you add another target not referenced ?

azayati (changes requested):
see review

azayati (approved):
LGTM

PR #373, opened 2021-12-13
Task-52255: Restore missed notification plugins labels

azayati:
Prior to this change and after cleaning useless news labels, the notification plugins labels are wrongly removed. With this PR, we will restore the missed notification plugins labels.


ahmadmido (approved):
LGTM after QA validation

PR #377, opened 2021-12-14
TASK-52070: Add role presentation for decorative images in news detail page

SaraBoutej:
In this PR we add role presentation for all user and space avatar, because we consider them as decorative images.


rdenarie:
@SaraBoutej Sonar make an alert for missing alt on the image you change.
Can you check if the alert is relevant or not

SaraBoutej:
> @SaraBoutej Sonar make an alert for missing alt on the image you change. Can you check if the alert is relevant or not

@rdenarie I made changes on avatar image that must be considered as a decorative image (with **no alt attribute** and by adding ro**le presentation**) to be ignored by Assistive Technology.

rdenarie:
So I remove the sonar alert

rdenarie (dismissed):

SaraBoutej:
> So I remove the sonar alert

@rdenarie can you please Reapprove

rdenarie (approved):

PR #382, opened 2021-12-20
Task-52286: Fix shared news activity link

hassenelkamel-exo:
Prior to this change, when a user is not member of a the space from which the news or news activity has been shared, from the stream, the shared article points to the original shared activity on which the user has not access. To solve this problem, we point the shared article instead to the share generated activity and not the original shared activity allowing the user to react on it.


azayati (approved):
LGTM after QA validation

PR #385, opened 2021-12-21
Task-52167: Fix cancel scheduling and drafts issues

hassenelkamel-exo:
Prior to this change, we have:
- The clock icon is displayed for a draft news from the news app 
- When we unschedule news, the scheduled date still exist and we can't reschedule the same news with the same scheduled date. 

After this change, the schedule problem is fixed by removing the property of schedule date when news is un scheduled and the clock icon displayed for a draft news from the news app is removed


azayati (dismissed):
LGTM after QA validation

azayati (approved):
LGTM after QA validation

PR #386, opened 2021-12-22
Task-51529: Fix archiving news

hassenelkamel-exo:
Prior to this change, archiving news doesn't work. To fix this problem, we ensure to call correctly the archive news function.


azayati (approved):
LGTM after QA validation

PR #390, opened 2021-12-23
Task-52417: Fix share news attachments

azayati:
Prior to this change, the shared in space permission on shared article attachment is not well set which make the attachment not shared. With this PR we fix the shared in space permission.


hassenelkamel-exo (approved):
LGTM

PR #396, opened 2021-12-29
MAINT-51975: Broadcast viewArticle event to analytics listener for already added viewers to allow count multiple views by user

hakermi:
…
**ISSUE**: When a user reads an article the event `viewArticle` is dispatched only when the user is never read the article, which doesn't allow to get the multiple views for a given user on an article in the analytics app
**FIX**: The fix will allow broadcasting the event `viewArticle` even if the user already added to viewers list in an article to allow save the action in the analytics app


ahamdi (changes requested):
Could you add a unit test for this change ?

hakermi:
> Could you add a unit test for this change ?

Unit test added

ahamdi (approved):

PR #400, opened 2022-01-03
Task-52602: Fix duplicated alert information message when sharing an article

hassenelkamel-exo:
Prior to this change, the alert information after sharing an article from news app is displayed twice. The problem is due to the call of the share activity `ExoNewsNotificationAlerts` component from news App. After this change, we ensure to call this component only when sharing activity from news apps.


azayati (approved):
LGTM after QA validation

PR #401, opened 2022-01-03
MAINT-51975: broadcast viewArticle event to analytics listener for already added viewers to allow count multiple views by user 

hakermi:
ISSUE: When a user reads an article the event viewArticle is dispatched only when the user is neved read the article, which doesn't allow to get the multiple views for a given user on an article in the analytics app
FIX: The fix will allow broadcasting the event viewArticle even if the user already added to viewers list in an article to allow save the action in the analytics app

MAINT-51975: Avoid markNewsAsRead trigger when getting article info in analytics app (#399)

ISSUE: The markAsRead function was placed in getNewsByid rest endpoint which will trigger this function when call the endpoint even from outside the news its self which is not correct
FIX: Create a new rest endpoint for marking news as read and call it when access the news details which seems more logic

### DIFF
--- a/services/src/main/java/org/exoplatform/news/service/impl/NewsServiceImpl.java
+++ b/services/src/main/java/org/exoplatform/news/service/impl/NewsServiceImpl.java
@@ -354,10 +354,8 @@ public class NewsServiceImpl implements NewsService {
   public void markAsRead(News news, String userId) throws Exception {
     if (!newsStorage.isCurrentUserInNewsViewers(news.getId(), userId)) {
       newsStorage.markAsRead(news, userId);
-      NewsUtils.broadcastEvent(NewsUtils.VIEW_NEWS, userId, news);
-    } else {
-      NewsUtils.broadcastEvent(NewsUtils.VIEW_NEWS, userId, news);
     }
+    NewsUtils.broadcastEvent(NewsUtils.VIEW_NEWS, userId, news);
   }
 
   /**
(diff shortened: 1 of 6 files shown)


services/src/main/java/org/exoplatform/news/service/impl/NewsServiceImpl.java
```diff
@@ -352,9 +352,11 @@ public News postNews(News news, String poster) throws Exception {
    */
   @Override
   public void markAsRead(News news, String userId) throws Exception {
-    if(!newsStorage.isCurrentUserInNewsViewers(news.getId(), userId)) {
+    if (!newsStorage.isCurrentUserInNewsViewers(news.getId(), userId)) {
       newsStorage.markAsRead(news, userId);
       NewsUtils.broadcastEvent(NewsUtils.VIEW_NEWS, userId, news);
+    } else {
+      NewsUtils.broadcastEvent(NewsUtils.VIEW_NEWS, userId, news);
     }
```
rdenarie:
```suggestion
    if (!newsStorage.isCurrentUserInNewsViewers(news.getId(), userId)) {
      newsStorage.markAsRead(news, userId);
    }
    NewsUtils.broadcastEvent(NewsUtils.VIEW_NEWS, userId, news);
```

rdenarie (changes requested):

eromdhaneexo (dismissed):

rdenarie (approved):

PR #405, opened 2022-01-07
Task-52908: Add ellipsis to latest news portlet name when it is too long

hassenelkamel-exo:
Prior to this change, when we add a portlet with a long name in the latest news template, it will not be well displayed. To fix the problem we ensure that the child component don't exceed the parent container.


azayati (approved):
LGTM after QA validation

PR #407, opened 2022-01-10
MAINT-52374: Fix xss in news content details

hakermi:
ISSUE: The news content has an xss vulnerability by using the v-html directive when displaying the news body, also the `HTMLSantizer` was doing its work and sanitize the body but the `StringEscapeUtils.unescapeHtml` was reverting the sanitization of the svg tag which caused an xss on svg script attributes.
FIX: Use v-html-sanitize instead to sanitize the news body and prevent svg dangerous attributes from being executed as a script also remove the unneeded `unescapeHtml` on the sanitized body


rdenarie (approved):
What about the news already stored ?
Does the content nee

hakermi:
> What about the news already stored ? Does the content nee

The sanitizer should treat this because it's launched when getting the news

PR #410, opened 2022-01-12
Task-52974: Fix edit publishing of shared articles

azayati:
Before this change, it is not possible to edit publishing options if a shared article, the json of shared in spaces field is not mapped correctly with rest endpoint entity.
After this change, we ensure to keep only space ids in the shared in spaces field in order to be mapped correctly.


hassenelkamel-exo (dismissed):
LGTM

hassenelkamel-exo (approved):
LGTM

PR #416, opened 2022-01-17
Task-52882: Don't display scheduled articles in news list portlets before being posted

ahmadmido:
Prior to this change, a scheduled article published in one or more targets is displayed in the corresponding news list portlets which is not logical, it should be displayed only when it is posted. After this change, when we associate the news to its target, the publication state is passed as a property which will allow to get only news with false staged property value.


services/src/main/java/org/exoplatform/news/service/NewsTargetingService.java
```diff
@@ -71,10 +87,11 @@
    * Save a {@link List} of {@link News} targets of a given {@link News} id by the current user
    *
    * @param newsId {@link News} identifier of {@link News} targets to be saved 
+   * @param isStaged {@link News} is staged news
    * @param targets {@link List} of {@link News} targets to be saved
    * @param currentUser current user attempting to save {@link News} targets
    */
-  void saveNewsTarget(String newsId, List<String> targets, String currentUser);
+  void saveNewsTarget(String newsId, boolean isStaged, List<String> targets, String currentUser);
```
azayati:
```suggestion
  void saveNewsTarget(String newsId, boolean staged, List<String> targets, String currentUser);
```

services/src/main/java/org/exoplatform/news/service/NewsTargetingService.java
```diff
@@ -71,10 +87,11 @@
    * Save a {@link List} of {@link News} targets of a given {@link News} id by the current user
    *
    * @param newsId {@link News} identifier of {@link News} targets to be saved 
+   * @param isStaged {@link News} is staged news
```
azayati:
```suggestion
   * @param staged {@link News} is staged
```

services/src/main/java/org/exoplatform/news/service/impl/NewsServiceImpl.java
```diff
@@ -402,7 +402,17 @@ public void publishNews(News newNews, String publisher) throws Exception {
...
+    }
+    //Update staged property to false when news is published.
+    Map<String, String> properties = new LinkedHashMap<>();
+    properties.put(PublicationDefaultStates.STAGED, String.valueOf(false));
+    List<MetadataItem> targets = newsTargetingService.getTargetsListByNewsId(newNews.getId());
+    if(!Objects.isNull(targets)) {
+      for (MetadataItem metadataItem : targets) {
+        metadataItem.setProperties(properties);
+        newsTargetingService.updateNewsTarget(metadataItem);
+      }
```
azayati: I think that you don't need this, you should just remove the above condition if(newNews.getTargets() != null && !oldTargets.equals(newNews.getTargets())) to delete and save news targets. You need that also to refresh the publish date

services/src/main/java/org/exoplatform/news/service/impl/NewsServiceImpl.java
```diff
@@ -402,7 +402,17 @@ public void publishNews(News newNews, String publisher) throws Exception {
     List<String> oldTargets = newsTargetingService.getTargetsByNewsId(newNews.getId());
     if(newNews.getTargets() != null && !oldTargets.equals(newNews.getTargets())) {
```
azayati:
```suggestion
    if(news.getTargets() != null && !oldTargets.equals(newNews.getTargets())) {
```

azayati (changes requested):
see review

azayati (approved):
LGTM

PR #424, opened 2022-01-25
MAINT-36140: Empty activity and server log errors when saving an article after deleting its in progress draft

hakermi:
…
**ISSUE**: When the user is about creating a new article and after filling some contents and that a draft is saved, when deleting the draft and then saving the article an empty activity is created while its related news node is deleted and no more exist, this leads also to some NPEs in log server.
**FIX**: This fix should handle this case by checking the existing of the news node and if it's deleted then, it will use the already filled content to recreate the news node and continue the post of the news correctly

### DIFF
--- a/services/src/main/java/org/exoplatform/news/service/impl/NewsServiceImpl.java
+++ b/services/src/main/java/org/exoplatform/news/service/impl/NewsServiceImpl.java
@@ -44,6 +44,8 @@ import org.exoplatform.social.metadata.model.MetadataItem;
 import org.exoplatform.social.notification.LinkProviderUtils;
 import org.exoplatform.upload.UploadService;
 
+import javax.jcr.ItemNotFoundException;
+
 /**
  * Service managing News and storing them in ECMS
  */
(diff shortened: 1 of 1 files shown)


services/src/main/java/org/exoplatform/news/service/impl/NewsServiceImpl.java
```diff
@@ -365,14 +367,22 @@ public News unScheduleNews(News news, org.exoplatform.services.security.Identity
...
-    sendNotification(poster, news, NotificationConstants.NOTIFICATION_CONTEXT.POST_NEWS);
-    if (news.isPublished()) {
-      publishNews(news, poster);
-    }
-    NewsUtils.broadcastEvent(NewsUtils.POST_NEWS_ARTICLE, news.getId(), news);//Gamification
-    NewsUtils.broadcastEvent(NewsUtils.POST_NEWS, news.getAuthor(), news);//Analytics
+    News existNews;
+    try {
+      existNews = newsStorage.getNewsById(news.getId(), false);
+    } catch (ItemNotFoundException e) {
```
rdenarie, line 373: as getNewsById throws a new Exception, are you sure that the type of the exception is ItemNotFound in this catch ?

hakermi: Yes i'm sure, and it's the exact exception that we should handle in this case

rdenarie (approved):

PR #426, opened 2022-01-26
Task-53138 : Fix scheduled articles problems

ahmadmido:
_ISSUES_ :
- After creating a scheduled article by a redactor or a manager and try to display it by another redactor or manager, the mention **By user** is displayed next to the scheduled time which is not the expected behavior.
- While opening a scheduled article, an error **Cannot read properties of null (reading 'id')** appears.
- When a user tries to display a not authorized scheduled article or an unpublished article posted in a space in which he is not a member via URL, undefined informations are displayed.
- When a scheduled article is posted, the mention updated on (+date of post ) is displayed even if the article is not updated.

_FIX_ :
- Display the mention **By user** only if a non scheduled article is updated.
- Prevent to display the `NewsActionFavorite` component in case of a scheduled article in addition and do not get the `newsId` value if the `news` object is passed as null from the parent component.
- A restriction page is displayed when the user tries to display a non unauthorized article.
- Prevent to display the updated date just after posting the scheduled article and display it only after the edit action.

### DIFF
--- a/webapp/src/main/webapp/news-details-app/components/ExoNewsDetailsApp.vue
+++ b/webapp/src/main/webapp/news-details-app/components/ExoNewsDetailsApp.vue
@@ -16,6 +16,7 @@
 </template>
 
 <script>
+const UNAUTHORIZED_CODE = 401;
 export default {
   props: {
     newsId: {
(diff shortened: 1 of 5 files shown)


azayati (approved):
LGTM after QA validation

PR #428, opened 2022-01-27
MAINT-36140: Empty activity and server log errors when posting a deleted draft article

hakermi:
ISSUE: When the user is about creating a new article and after filling some contents and that a draft is saved, when deleting the draft and then posting the article an empty activity is created. This leads also to some NPEs in log server.
FIX: This fix should handle this case by checking the existing of the news node and if it's deleted then, it will use the already filled content to recreate the news node and continue the post of the news correctly.


services/src/main/java/org/exoplatform/news/service/impl/NewsServiceImpl.java
```diff
@@ -365,14 +367,22 @@ public News unScheduleNews(News news, org.exoplatform.services.security.Identity
...
-      publishNews(news, poster);
-    }
-    NewsUtils.broadcastEvent(NewsUtils.POST_NEWS_ARTICLE, news.getId(), news);//Gamification
-    NewsUtils.broadcastEvent(NewsUtils.POST_NEWS, news.getAuthor(), news);//Analytics
+    News existNews;
+    try {
+      existNews = newsStorage.getNewsById(news.getId(), false);
+    } catch (ItemNotFoundException e) {
+      existNews = newsStorage.createNews(news);
+    }
```
azayati: this method is dedicated only to post a news when it is already created. you should fix the issue on the create method cc @rdenarie

azayati (changes requested):
see review

services/src/main/java/org/exoplatform/news/service/impl/NewsServiceImpl.java
hakermi: @azayati the problem is the post and not in the create method, when post a news already deleted before the creation  this fix should be okay
what i should fix in the create function? the news created but it's deleted by the user before it's post while still in the edit mode,
This fix will use the content already filled and recreate the news and continue the post normally

azayati: create method calls post method, and here you call create method from post method. It is a cyclic call which is not permitted. You can check the existence of the news from create method before calling the post method

hakermi: > create method calls post method, and here you call create method from post method. It is a cyclic call which is not permitted. You can check the existence of the news from create method before calling the post method

i'm calling the create method of the storage to recreate the news node 
https://github.com/exoplatform/news/blob/develop/services/src/main/java/org/exoplatform/news/service/impl/NewsServiceImpl.java#L107

hakermi: @azayati Okay i understood your point

rdenarie: As I understand, we should rework the pr.
Once done, we should retest it. So please add your modification on feature/maitenance, so that QA will be able to retest

hakermi: @azayati Could you check now ?

services/src/main/java/org/exoplatform/news/service/impl/NewsServiceImpl.java
```diff
@@ -104,7 +106,14 @@ public News createNews(News news, org.exoplatform.services.security.Identity cur
...
+        try {
+          createdNews = newsStorage.getNewsById(news.getId(), false);
+        } catch (ItemNotFoundException e) {
+          createdNews = newsStorage.createNews(news);
+        }
+        if (createdNews != null) {
+          createdNews = postNews(news, currentIdentity.getUserId());
+        }
       } else if (news.getSchedulePostDate() != null){
         createdNews = unScheduleNews(news, currentIdentity);
```
azayati, line 111: @hakermi normally with the same scenario we will also a server exception when we try to unschedule a deleted draft article, so we should fix that also

azayati (changes requested):
see review

services/src/main/java/org/exoplatform/news/service/impl/NewsServiceImpl.java
hakermi: Ok

azayati (approved):
LGTM

PR #429, opened 2022-01-27
Task-52284: Fix Back button on news articles to lead to last viewed page

abenmbarek:
*ISSUE* : back button on news article details always leads to the AS in case of the user is a member and if he's not, it leads him to the snapshot page from anywhere he opens the article, because of a regression occurred somehow.

*FIX* : fix the regression by re-adding the `goBack` function which redirect to the last visited page. If the user is using a new tab in case the history navigation is clear, the go back button wont be displayed.


azayati (approved):
LGTM after QA validation

PR #431, opened 2022-01-28
Task-52284: Fix Back button on news articles to lead to previous viewed page

abenmbarek:
_ISSUE_ : back button on news article details always leads to the AS in case of the user is a member and if he's not, it leads him to the snapshot page from anywhere he opens the article, because of a regression occurred somehow.

_FIX_ : fix the regression by re-adding the `goBack` function which redirect to the last visited page. If the user is using a new tab in case the history navigation is clear, the go back button wont be displayed.

### DIFF
--- a/webapp/src/main/webapp/news-details/components/ExoNewsDetailsToolBar.vue
+++ b/webapp/src/main/webapp/news-details/components/ExoNewsDetailsToolBar.vue
@@ -1,6 +1,6 @@
 <template>
   <div :class="!showEditButton && 'me-5'" class="newsDetailsTopBar">
-    <a class="backBtn" :href="backURL"><i class="uiIconBack my-4"></i></a>
+    <a class="backBtn" @click="goBack"><i class="uiIconBack my-4"></i></a>
     <v-btn
       v-if="publicationState === 'staged'"
       class="btn newsDetailsActionMenu mt-6 mr-2 pull-right"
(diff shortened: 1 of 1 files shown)


webapp/src/main/webapp/news-details/components/ExoNewsDetailsToolBar.vue
```diff
@@ -87,7 +87,19 @@ export default {
...
       return this.news && this.news.published;
+    },
+    lastVisitedPage(){
+      return history && history.length && history.length > 2;
     }
   },
+  methods: {
+    goBack() {
+      if (this.lastVisitedPage)
+      {history.back();}
```
azayati:
```suggestion
      if (this.lastVisitedPage) {
        history.back();
      }
```

webapp/src/main/webapp/news-details/components/ExoNewsDetailsToolBar.vue
```diff
@@ -73,7 +73,7 @@ export default {
     };
   },
   computed: {
-    backURL() {
+    backUrlHistoryCleared() {
```
azayati:
```suggestion
    historyClearedBackUrl() {
```

azayati (changes requested):
see review

azayati (approved):
LGTM after QA validation

PR #433, opened 2022-02-02
Task-53648: Fix displaying news update date in news app for all languages

azayati:
Prior to this change, when to switching the current language to some languages like Spanish, Chinese .. the news update date is not well formatted with the current locale using toLocaleDateString(locale, options) function which causes a problem to display news items from news app. After this change, we ensure to use date-format vue component to display the news update date which ensure formatting the date automatically to the current locale format.


webapp/src/main/webapp/news/components/NewsAppItem.vue
```diff
@@ -74,7 +74,12 @@ along with this program. If not, see <http://www.gnu.org/licenses/>.
               :format="dateTimeFormat"
               class="newsTime" />
           </span>
-          <span v-else>{{ news.updatedDate }}</span>
+          <span v-else>
+            <date-format
+              :value="news.updatedDate"
+              :format="dateFormat"
+              class="newsTime" />
+          </span>
```
abenmbarek, line 82:
```suggestion
          <span v-else>
            <date-format
              :value="news.updatedDate"
              :format="dateFormat"
              class="newsTime" />
                          -
            <date-format
              :value="news.updatedDate"
              :format="dateTimeFormat"
              class="newsTime" />
          </span>
```

abenmbarek (changes requested):

webapp/src/main/webapp/news/components/NewsAppItem.vue
azayati: Initially we displayed the updated date without time, it is different from scheduled date for which schedule time is needed. So we keep the displaying of updated date as is.

abenmbarek (approved):
LGTM after QA validation

PR #438, opened 2022-02-07
Task-49608 News not displayed on snapshot view

rdenarie:
Before this fix there was a js error when latestNews initialize. The error is that Vue.createApp does not exists.
This problem is due to a missing dependency on js module commonVueComponent, which add the function createApp in Vue object
This commit add the missing dependency in gatein-resources.xml


montassarhamadi (approved):

rdenarie:
Do not merged this currently. We are waiting the release of 6.2.4 before merging

wassimmhamdi (approved):

PR #439, opened 2022-02-08
Task-53093: Add news list portlet in content application category for full data instances

azayati:
Prior to this change, news list portlet is not added into content application category for full data instances and added only for empty instances. After this change, we have configured 'system' value param on ApplicationCategoriesPlugins component plugin to allow that.


hassenelkamel-exo (approved):
LGTM

PR #440, opened 2022-02-08
Task-54069: Display all targets when publishing articles

hassenelkamel-exo:
Before this change, target selector of publishing drawer displays only referenced targets. So the newly created targets are not selectable from the publishing drawer when they are not yet referenced by one or more news list portlets instances. After this change, and waiting the update of referenced targets when any page modification is done, we will display all targets from publishing drawer.


azayati (dismissed):
LGTM after QA validation

azayati (approved):
LGTM

PR #455, opened 2022-02-16
Task-53818: Fix loading problem of news targets management page

hassenelkamel-exo:
**Issue:**  When accessing to news targets management page, the page still loads without displaying news targets.
**Fix:**  We define newsTargetingService in the main.js file of news Publish Targets Management portlet to ensure being loaded when the portlet is displayed.


azayati (changes requested):
Add a description of the problem and the fix @hassenelkamel-exo

azayati (dismissed):
LGTM after QA validation

azayati (approved):
LGTM

PR #475, opened 2022-03-14
Task-55137: Do not display archived articles in the latest news

ahamdi:
Prior to this change, when post an article with title and content, then publish it with choosing Snapshot latest news as target
,click to Archive button related to this article , then go to the Snapshot page, Article is displayed .
This is due to the fact that article displayed in the latest news section is not filtered by not archived ,
After this change, we ensure that article archived is not displayed in the latest news section of the snapshot page.

### DIFF
--- a/packaging/pom.xml
+++ b/packaging/pom.xml
@@ -24,7 +24,7 @@
   <parent>
     <artifactId>news</artifactId>
     <groupId>org.exoplatform.addons.news</groupId>
-    <version>2.3.x-SNAPSHOT</version>
+    <version>2.3.x-maintenance-SNAPSHOT</version>
   </parent>
   <artifactId>news-packaging</artifactId>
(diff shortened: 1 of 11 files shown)


rdenarie (changes requested):
Filtering the news after applying the limit and the offset can lead to a display problem :

For example, if you have 10 news, news1, news2, news3, news4 are archived, then you make the request to get news with offset 0 and limit 5, the service will get [news1, news2, news3, news4, news5], then you apply filter, and then you display only news5, instead of displaying 5 news : news5, news6, news7, news8, news9

cc @ahamdi @azayati What do you think ? should we update the rest service to add an option for saying if we want archived news or not ?

ahamdi:
@hchaabene Could you take care of the feedback of @rdenarie  ?

PR #478, opened 2022-03-17
Task-55322: Confirm button is not disabled when Edit the scheduling post later when chose date and time

hchaabene:
Prior to this change, when add User as redactor to space page,  user connect to PLF and go to space , click to space then write an article , add title and content then click on post , click post later and click Schedule then click on Scheduled article
This is due to the fact that button is enabled but date or time  not change in Schedule article ,
After this change, we ensure that button is disabled without changes at date or time.


webapp/src/main/webapp/schedule-news-drawer/components/ExoScheduleNewsDrawer.vue
```diff
@@ -258,6 +258,7 @@ export default {
     },
   },
   data: () => ({
+    isDisabledTime: false,
```
rdenarie: To remove

rdenarie (changes requested):
We still dont know why, when the drawer opens, the function postDate is called with, as newVal, the current time.
Since we dont understand why we have this, we will not able to fix correctly the problem.
You need to dig this point, to understand what is the current behaviour. Then, once we exactly now that, we will be able to find a solution

webapp/src/main/webapp/schedule-news-drawer/components/ExoScheduleNewsDrawer.vue
```diff
@@ -274,12 +274,20 @@ export default {
   }),
   watch: {
     postDate(newVal, oldVal) {
+      console.log('newVal',newVal ,' oldVal ',new Date(newVal).getTime() );
```
rdenarie: Please remove console.log

webapp/src/main/webapp/schedule-news-drawer/components/ExoScheduleNewsDrawer.vue
```diff
@@ -274,12 +274,20 @@ export default {
...
       const scheduleDate = new Date(this.schedulePostDate);
-      this.disabled = postDate.getTime() === scheduleDate.getTime();
+      
+      const postDateString = postDate.getFullYear().toString() + postDate.getMonth().toString()+ postDate.getDate().toString() ;
+      const scheduleDateString = scheduleDate.getFullYear().toString() + scheduleDate.getMonth().toString() + scheduleDate.getDate().toString();
+      
+      const postTimeString = postDate.getHours().toString() + postDate.getMinutes().toString();
+      const scheduleTimeString = scheduleDate.getHours().toString() + scheduleDate.getMinutes().toString();
+
+      this.disabled = ( (  postDateString ===  scheduleDateString ) && ( postTimeString  === scheduleTimeString  ) ) ;
```
rdenarie: can you format this line ? there is a lot of trailling spaces to remove

rdenarie (changes requested):

rdenarie (approved):

PR #479, opened 2022-03-17
Task-55240: Disable write an article feature when a poll is created

abenmbarek:
Prior to this change, we can write an article even if a poll is created and not yet posted. After this change, we will disallow writing an article when a poll is created.


azayati (approved):
LGTM after QA validation

PR #482, opened 2022-03-28
Task-55722 : Fix exception when liking or commenting an activity

rdenarie:
Prior to this change, when liking or commenting any activity, exceptions coming from NewsActivity like and comment listeners are raised. After this commit, we ensure to execute these listeners only for news activity type.


azayati (approved):
LGTM

PR #496, opened 2022-04-13
Task-55491: Implement News cards slider template view component

abenmbarek:
prior to this change, we will be able to choose a new way to display the articles by using the template "Cards slider".


azayati (approved):
LGTM

PR #497, opened 2022-04-13
TASK-55664 Upgrade ESLint version to use newer ECMAScript version

boubaker:
Upgrade ESLint version to use new ECMAScript 2022.


abenmbarek (approved):
Using the optional chaining is so helpful and easy instead of typing too long code lines

PR #498, opened 2022-04-14
Task-56259: Fix news slider template height

azayati:
Prior to this change, when a news list portlet is added outside slider container of snapshot page and configured with news slider template, the height is set to the default implicit value 500px which is not the adequate one. After this fix, the height is set explicitly to 220px.


abenmbarek (approved):
LGTM after QA validation

PR #505, opened 2022-04-25
Task-56520: Display the author name in published article detail of a hidden space

abenmbarek:
Prior this fix, the author name of a published article in a hidden space cannot be displayed for a non-member. After this fix, a non-member of a hidden space can see the published article's author name.


azayati (approved):
LGTM

PR #506, opened 2022-04-26
Task-56520: Dislpay the author name in case of hidden space and published article

abenmbarek:
Prior this fix, the author name of a published article in a hidden space cannot be displayed for a non-member. After this fix, a non-member of a hidden space can see the published article's author name only of the `news` object is not null.


azayati (approved):
LGTM

boubaker (approved):

PR #509, opened 2022-04-28
Task 56744: Search by favorites and news (article bookmarked/unbookmarked from the activity )

montassarhamadi:
Problem: news still displays in search by using favorite filtering after unbookmarking.
Fix: the process of deleting activity was completed from the favorite list but not the case for activity details (news) because there is a comparison returned boolean false and ignored the deleting of activity details. that's why the details stay saved and then displayed in search and also in recent favorites.


ahamdi (changes requested):
Could you add a UT for this change ?

montassarhamadi:
> Could you add a UT for this change ?

done

ahamdi (changes requested):
We need a test for removing a favorite news

Montassar-Hamadi:
> We need a test for removing a favorite news

I have no idea about UT this is the first time when

> We need a test for removing a favorite news

It already exists @ahamdi in module social. check this link : 
https://github.com/Meeds-io/social/blob/e6fdfc36e394a890106b9d125d554b731c801584/component/core/src/test/java/org/exoplatform/social/metadata/favorite/FavoriteServiceTest.java#L133

montassarhamadi:
> We need a test for removing a favorite news

It already exists @ahamdi in module social. check this link :

https://github.com/Meeds-io/social/blob/e6fdfc36e394a890106b9d125d554b731c801584/component/core/src/test/java/org/exoplatform/social/metadata/favorite/FavoriteServiceTest.java#L133

ahamdi:
Delete operation is not covered by the UT of the listener

ahamdi (approved):

PR #510, opened 2022-04-28
Task-56521: Fix duplication of news and activity items in favorites drawer

SaraBoutej:
Before this Fix when we bookmark a news, we have the related activity that is bookmarked too, so in the drawer of recent favorites we have duplication of items: activity + news.
With this fix, we add a boolean property "hidden" when bookmark an item to avoid the duplication of an item in the favorites drawer.


webapp/src/main/webapp/news-details/components/ExoNewsFavoriteAction.vue
```diff
@@ -65,7 +66,7 @@ export default {
     },
     added() {
       this.displayAlert(this.$t('Favorite.tooltip.SuccessfullyAddedAsFavorite', {0: this.$t('news.label')}));
-      this.$favoriteService.addFavorite('news', this.news.id)
+      this.$favoriteService.addFavorite('news', this.news.id, '', true)
```
boubaker, line 69:
```suggestion
      this.$favoriteService.addFavorite('news', this.news.id, null, true)
```

webapp/src/main/webapp/news-details/components/ExoNewsFavoriteAction.vue
```diff
@@ -2,6 +2,7 @@
   <favorite-button
     :id="activityId"
     :favorite="isFavorite"
+    :hidden=true
```
boubaker, line 5:
```suggestion
    hidden
```
You can just add the name of the attribute to change its value to `true`
See https://v2.vuejs.org/v2/guide/components-props.html#Passing-a-Boolean

PR #514, opened 2022-04-29

abenmbarek:
Prior to this change, there is no check of news objects nullability before displaying it in news list portlets. After this fix, we added nullability checks to avoid displaying null articles.


azayati (approved):
LGTM

PR #515, opened 2022-04-29

abenmbarek:
Prior to this change, there is no check of null news objects before displaying into templates.
After this fix, we added some nullable checks to avoid any display of null objects from the templates.

### DIFF
--- a/webapp/src/main/webapp/news-list-view/components/NewsListView.vue
+++ b/webapp/src/main/webapp/news-list-view/components/NewsListView.vue
@@ -175,7 +175,7 @@ export default {
       this.loading = true;
       return this.$newsListService.getNewsList(this.newsTarget, this.offset, this.limit, true)
         .then(newsList => {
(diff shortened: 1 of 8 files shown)


azayati (dismissed):
LGTM

webapp/src/main/webapp/news-list-view/components/NewsListView.vue
```diff
@@ -175,7 +175,7 @@ export default {
       this.loading = true;
       return this.$newsListService.getNewsList(this.newsTarget, this.offset, this.limit, true)
         .then(newsList => {
-          this.newsList = newsList.news || [];
+          this.newsList = newsList.news.filter(news => news !== null) || [];
```
ahamdi:
```suggestion
          this.newsList = newsList.news.filter(news => !!news ) || [];
```

ahamdi (approved):

PR #519, opened 2022-04-29

abenmbarek:
Prior to this change, there is no check of null news objects before displaying into templates.
After this fix, we added some nullable checks to avoid any display of null objects from the templates.


ahamdi (approved):

azayati (approved):
LGTM

PR #520, opened 2022-05-06
Task-56942: Accents are not supported on news targets name

AzmiTouil:
Prior to this change, we can't add a news targets with a name that contain accented character.
Fix: Update Reg expression using the unicode values for accented characters.


boubaker (approved):
I added a comment on Task. In fact, that's weird to ad this validation on title of news. I validate this to unblock this specific use case, but we should reconsider this Use Cases and challenge the POs IMO.

PR #525, opened 2022-05-09
TASK-56990: Wrong last modifier value in news drafts management

hakermi:
Prior to this change, when a user (redactor) creates a news article in a space and then modify its content without saving to keep a draft, at this case when another redactor edits the article a warning message appears to tell him that he is about to edit a draft created by user 1, but instead it says that it's a draft created by user x which is the last user reader and not editor. this due to the fact that the `exo:lastModifier` is updating when add , remove or update a property in the jcr node which is the case  when mark a news as read, so the username of the reader will be set in th `exo:lastModifier` value.
This PR adds a new mixin called `exo:newsModification` with an `exo:newsLastModifier` property, which will be updated when update the news and not when marking it as read.


boubaker (dismissed):

boubaker:
please resolve Sonar issue first

boubaker (approved):

PR #528, opened 2022-05-10
Task-57041: Adjust news list display in small containers and mobile

abenmbarek:
Prior to this fix, when putting the news list template in a small container or using mobile device, the display is wrong: only images are shown and listed in a row.
After this fix, the news list template will be well displayed vertically, using flex wrap, in case of a small container or a mobile device.


azayati (approved):
LGTM

PR #530, opened 2022-05-10
TASK-57050: No notification is sent to the mentioned user when the mention is done from the edition form

IlhemEssaadi:
ISSUES : When updating an article and mentioning a user, no notification is sent to the mentioned user.
FIX : The problem is that the mentioned users have been removed from the body of the updated news before sending them in the notification. We fixed that by adding to the news object an originalBody attribute containing the body of the original news from  which we get the previous mentions and remove them from the updated news.

### DIFF
--- a/services/src/main/java/org/exoplatform/news/service/impl/NewsServiceImpl.java
+++ b/services/src/main/java/org/exoplatform/news/service/impl/NewsServiceImpl.java
@@ -151,7 +151,8 @@ public class NewsServiceImpl implements NewsService {
     if (!canEditNews(news, updater)) {  
       throw new IllegalArgumentException("User " + updater + " is not authorized to update news");
     }
-    Set<String> previousMentions = NewsUtils.processMentions(news.getBody());
+    News originalNews = newsStorage.getNewsById(news.getId(), false);
+    Set<String> previousMentions = NewsUtils.processMentions(originalNews.getOriginalBody());
     if (publish != news.isPublished() && news.isCanPublish()) {
       news.setPublished(publish);
       if (news.isPublished()) {
--- a/services/src/main/java/org/exoplatform/news/model/News.java
+++ b/services/src/main/java/org/exoplatform/news/model/News.java
@@ -20,7 +20,11 @@ public class News {
 
   private String               summary;
 
+  /* sanitizedBody with usernames */
   private String               body;
+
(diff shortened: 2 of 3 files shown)


services/src/main/java/org/exoplatform/news/service/impl/NewsServiceImpl.java
```diff
@@ -173,14 +172,7 @@ public News updateNews(News news, String updater, Boolean post, boolean publish)
...
       // Send mention notifs
       if (StringUtils.isNotEmpty(news.getId()) && news.getCreationDate() != null) {
-        News newMentionedNews = news;
-        if (!previousMentions.isEmpty()) {
-          //clear old mentions from news body before sending a custom object to notification context.
-          previousMentions.forEach(username -> {
-            newMentionedNews.setBody(newMentionedNews.getBody().replaceAll("@"+username, ""));
-          });
-        }
-        sendNotification(updater, newMentionedNews, NotificationConstants.NOTIFICATION_CONTEXT.MENTION_IN_NEWS);
```
azayati: Normally you should keep this treatment in order to not send additional notifications to the users already mentioned.
The problem is around this instruction `Set<String> previousMentions = NewsUtils.processMentions(news.getBody());`

- Here news is the new updated news object

You should get instead the old news object with `newsStorage.getNewsById(news.getId(), false)`

cc @boubaker

azayati (changes requested):
see review

services/src/main/java/org/exoplatform/news/service/impl/NewsServiceImpl.java
```diff
@@ -151,7 +151,8 @@ public News updateNews(News news, String updater, Boolean post, boolean publish)
     if (!canEditNews(news, updater)) {  
       throw new IllegalArgumentException("User " + updater + " is not authorized to update news");
     }
-    Set<String> previousMentions = NewsUtils.processMentions(news.getBody());
+    News previousNews = newsStorage.getNewsById(news.getId(), false);
```
azayati:
```suggestion
    News originalNews = newsStorage.getNewsById(news.getId(), false);
```

azayati (changes requested):
see review

azayati (approved):
LGTM

PR #533, opened 2022-05-11
TASK-56521 Avoid duplicating Note Favorite & Tag Metadatas with Activity

boubaker:
Prior to this change, when adding a news as favorite from Activity stream, a Favorite Metadata is associated to Activity and to the News at the same time (Same for tags). This fix will ensure to associate the favorite Metadata to the News content only. In fact, an improvement for Activity Metadata Management has been added (https://github.com/Meeds-io/social/pull/1460) that allows to add a specific target Object Type+Id when displaying activity in Stream. Thus, the Favorite button will manage the favorites of News Content Object identified by its real id.


AzmiTouil (approved):
LGTM

PR #542, opened 2022-05-16
Task-57193: News activity update mail notifications should be sent only for non draft status

sofyenne:
ISSUE : When a user creates an article in a space and another user likes or comments it. If the creator edits the article, the liker or commenter receives many mail notifications.
FIX : The news activity update should be done only when the article hasn't draft status.


azayati (approved):
LGTM, FYI @sofyenne I have modified the PR title and description to describe properly the issue and fix. To  be tested also for onsite notifications

PR #546, opened 2022-05-17
TASK-56875: Fix external user does not receive published news notification

IlhemEssaadi:
ISSUES : When we post and publish news, the external user does not receive published news notification .
FIX : The problem that the notification of the published news are sent to the internal users, so it fixed by getting the receivers that are the members of the space (internals and externals).


azayati (changes requested):
The right excepted behavior should be verified with the PO because the last [PR](https://github.com/exoplatform/news/pull/327) is the opposite of what is asked in this US.

azayati (commented):
to be closed @IlhemEssaadi since confirmed with PO and QA that it is an expected behavior and not a bug

PR #550, opened 2022-05-18
Task-56827: Add text truncate in mobile when space name is too long

lbengayed:
Prior to this change, When the space name is too long, the first letter should be displayed.
Fix: Add max-width to the space name

### DIFF
--- a/webapp/src/main/webapp/skin/less/news.less
+++ b/webapp/src/main/webapp/skin/less/news.less
@@ -3515,7 +3515,6 @@
                 margin-left: 5px ~'; /** orientation=lt */ ';
                 margin-right: 5px ~'; /** orientation=rt */ ';
                 align-items: center;
-                max-width: 60px;
 
                 a {
                   margin-left: 5px ~'; /** orientation=lt */ ';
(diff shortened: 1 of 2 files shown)


webapp/src/main/webapp/skin/less/news.less
```diff
@@ -3515,7 +3515,7 @@
                 margin-left: 5px ~'; /** orientation=lt */ ';
                 margin-right: 5px ~'; /** orientation=rt */ ';
                 align-items: center;
-                max-width: 50px;
+                max-width: 60px;
```
boubaker: To make it responsive without fixed width, it's better to add `text-truncate` in https://github.com/exoplatform/news/blob/aea484882a923cc5c3a0f352773dc16978a6016d/webapp/src/main/webapp/news/components/NewsAppItem.vue#L59-L65
and delete this `max-width`
```suggestion
```

lbengayed: @boubaker I tested it from the beginning with text-truncate and it does not solve the problem

boubaker: https://user-images.githubusercontent.com/594824/169070022-116e6f7b-7a37-487e-a4aa-a9ced4d5c3d6.mp4

boubaker: I let @hghorbel confirm the expected behavior

boubaker (approved):

PR #552, opened 2022-05-18
Task-56676: A space member should be allowed to Bookmark a news

montassarhamadi:
Problem: Favorite icon is not displayed at the right side on the news details for member in space X.
Fix: the display of Favorite icon will be made independently from edit permission for current user.


boubaker (approved):
Please use new title and description while merging

PR #555, opened 2022-05-19
Task-57267: Fix translation of update button

hassenelkamel-exo:
Before the change, update button is not translated when changing platform language.


hakermi (approved):

azayati (approved):
LGTM

PR #559, opened 2022-05-20
Task-54409: Fix displaying numbers of views and date for posted news with draft status

sofyenne:
ISSUE : When the user cancel editing a news the numbers of views and the date are not displayed. 
FIX : Display the numbers of views and the date .

### DIFF
--- a/webapp/src/main/webapp/news/components/NewsAppItem.vue
+++ b/webapp/src/main/webapp/news/components/NewsAppItem.vue
@@ -65,7 +65,7 @@ along with this program. If not, see <http://www.gnu.org/licenses/>.
             small-font-size
             popover />
         </div>
-        <div v-if="!draftNews" class="newsDate pe-4">
+        <div class="newsDate pe-4" v-if="news.activityId">
           <i v-if="displayClock" class="uiIconClock"></i>
           <span v-if="news && news.schedulePostDate">
             <date-format
@@ -85,7 +85,7 @@ along with this program. If not, see <http://www.gnu.org/licenses/>.
               class="newsTime caption" />
           </span>
         </div>
-        <div class="newsViews " v-if="!news.draft && !news.scheduled">
+        <div class="newsViews " v-if="!news.scheduled && news.activityId">
           <i class="uiIconWatch"></i>
           <span class="viewsCount">{{ news.viewsCount }}  {{ $t('news.app.views') }}</span>
         </div>


webapp/src/main/webapp/news/components/NewsAppItem.vue
```diff
@@ -65,7 +65,7 @@ along with this program. If not, see <http://www.gnu.org/licenses/>.
             small-font-size
             popover />
         </div>
-        <div v-if="!draftNews" class="newsDate pe-4">
+        <div class="newsDate pe-4">
```
azayati: You should test this modification for a draft article not already posted, normally these informations should not be displayed in that case. (otherwise to be verified with the PO)

webapp/src/main/webapp/news/components/NewsAppItem.vue
```diff
@@ -85,7 +85,7 @@ along with this program. If not, see <http://www.gnu.org/licenses/>.
               class="newsTime caption" />
           </span>
         </div>
-        <div class="newsViews " v-if="!news.draft && !news.scheduled">
+        <div class="newsViews " v-if="!news.scheduled">
```
azayati: You should test this modification for a draft article not already posted, normally these informations should not be displayed in that case. (otherwise to be verified with the PO)

azayati (changes requested):
see review

webapp/src/main/webapp/news/components/NewsAppItem.vue
azayati: FYI in order to distinguish between a draft article not already posted and another draft of a posted article, you can base on news.activities which will be null in the first case and not null in the second case

azayati: IMO, you need to add the condition news.activityId also here, WDYT ?

azayati (changes requested):

webapp/src/main/webapp/news/components/NewsAppItem.vue
```diff
@@ -65,7 +65,7 @@ along with this program. If not, see <http://www.gnu.org/licenses/>.
             small-font-size
             popover />
         </div>
-        <div v-if="!draftNews" class="newsDate pe-4">
+        <div class="newsDate pe-4" v-if="news && news.activityId">
```
abenmbarek, line 68: tyhty

abenmbarek:
```suggestion
        <div class="newsDate pe-4" v-if="news?.activityId">
```

webapp/src/main/webapp/news/components/NewsAppItem.vue
```diff
@@ -85,7 +85,7 @@ along with this program. If not, see <http://www.gnu.org/licenses/>.
               class="newsTime caption" />
           </span>
         </div>
-        <div class="newsViews " v-if="!news.draft && !news.scheduled">
+        <div class="newsViews " v-if="!news.scheduled && news.activityId">
```
abenmbarek, line 88:
```suggestion
        <div class="newsViews " v-if="!news?.scheduled && news?.activityId">
```

azayati (approved):
LGTM

PR #581, opened 2022-05-31
Task-57143: Adjust display of news headlines template using a  small width container

abenmbarek:

### DIFF
--- a/webapp/src/main/webapp/news-list-view/components/views/NewsLatestView.vue
+++ b/webapp/src/main/webapp/news-list-view/components/views/NewsLatestView.vue
@@ -15,17 +15,44 @@ You should have received a copy of the GNU Affero General Public License
 along with this program. If not, see <http://www.gnu.org/licenses/>.
 -->
 <template>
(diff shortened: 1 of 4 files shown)


webapp/src/main/webapp/news-list-view/components/views/NewsLatestView.vue
```diff
@@ -83,6 +110,7 @@ export default {
   },
   mounted() {
     this.$nextTick().then(() => this.$root.$emit('application-loaded'));
+    this.hasSmallWidthContainer = (this.$refs.newsLatestView?.clientWidth *100 / window.screen.width) < 33;
```
ahamdi, line 113: Why are you using this check ? couldn't we use Mobile vs Tablet vs Desktop?

hchaabene: this method allows to detect the dimension of the screen and to check if the display of the mosaic template is by column or by the normal form.

PR #582, opened 2022-05-31
Task-57165: News mosaic template is ko if small width

hchaabene:
Prior to this change, first publish 4 articles with names not truncated  , when i move latest new template in a small container or using mobile device to the right side or in a column , then  choose mosaic model , 
This is due to changing the width is small, the display is bad and names of articles are truncated.
After this change,  the article is readable , and names of articles are not truncated.

### DIFF
--- a/webapp/src/main/webapp/news-list-view/components/views/NewsMosaicView.vue
+++ b/webapp/src/main/webapp/news-list-view/components/views/NewsMosaicView.vue
@@ -15,12 +15,12 @@ You should have received a copy of the GNU Affero General Public License
 along with this program. If not, see <http://www.gnu.org/licenses/>.
 -->
 <template>
-  <div id="top-news-mosaic">
+  <div id="top-news-mosaic" ref="top">
     <div :class="`mosaic-container ma-3 ${smallHeightClass}`">
       <div 
         v-for="(item, index) of news"
         :key="index"
-        class="article"
+        :class="isSmallWidth ? 'articleSmallWidth' : 'article'"
         :id="`articleItem-${index}`">
         <a
           class="articleLink"
@@ -76,6 +76,7 @@ export default {
         month: 'long',
         day: 'numeric',
       },
+      isSmallWidth: false
     };
   },
   created() {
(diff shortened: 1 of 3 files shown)


webapp/src/main/webapp/news-list-view/components/views/NewsMosaicView.vue
```diff
@@ -76,20 +87,24 @@ export default {
...
+      isSmallWidth: false
     };
   },
   created() {
     this.reset();
     this.$root.$on('saved-news-settings', this.refreshNewsViews);
     this.getNewsList();
   },
+  updated() {
+    this.isSmallWidth =  this.$refs.top?.clientWidth *100 / window.screen.width  < 33;
```
ahamdi: Why are you using this check ? couldn't we use Mobile vs Tablet vs Dersktop?

hchaabene: this method allows to detect the dimension of the screen and to check if the display of the mosaic template is by column or by the normal form.

ahamdi (dismissed):

webapp/src/main/webapp/news-list-view/components/views/NewsMosaicView.vue
```diff
@@ -15,12 +15,12 @@ You should have received a copy of the GNU Affero General Public License
 along with this program. If not, see <http://www.gnu.org/licenses/>.
 -->
 <template>
-  <div id="top-news-mosaic">
+  <div id="top-news-mosaic" ref="top">
```
ahamdi:
```suggestion
  <div id="top-news-mosaic" ref="top-news-mosaic">
```

webapp/src/main/webapp/news-list-view/components/views/NewsMosaicView.vue
```diff
@@ -76,20 +76,24 @@ export default {
...
+      isSmallWidth: false
     };
   },
   created() {
     this.reset();
     this.$root.$on('saved-news-settings', this.refreshNewsViews);
     this.getNewsList();
   },
+  mounted() {
+    this.isSmallWidth =  this.$refs.top?.clientWidth *100 / window.screen.width  < 33;
```
ahamdi: Could you adjust the reference ID here too ?

webapp/src/main/webapp/news-list-view/components/views/NewsMosaicView.vue
```diff
@@ -102,6 +106,9 @@ export default {
           .finally(() => this.initialized = false);
       }
     },
+    minLength(lengthNews){
+      return lengthNews<5 && lengthNews>0 ? 100/lengthNews : 25;
```
ahamdi:
```suggestion
      return lengthNews < 5 && lengthNews > 0 ? 100 / lengthNews : 25;
```

ahamdi (approved):

PR #587, opened 2022-06-03
TASK-51498: Fix Copy and paste a document from Google Doc into an article

IlhemEssaadi:
ISSUES : When we copy a text from a google Doc and paste it in article ,The content is bolded automatically.
FIX : add googleDocPastePlugin plugin to extraPlugins of ckeditor which allows pasting content from google docs to ckeditor in the same format.


ahamdi (changes requested):
Wait before merging

PR #588, opened 2022-06-07
Task-57143: Adjust display of news headlines template using a small width container

hchaabene:

### DIFF
--- a/webapp/src/main/webapp/news-list-view/components/views/NewsLatestView.vue
+++ b/webapp/src/main/webapp/news-list-view/components/views/NewsLatestView.vue
@@ -15,12 +15,12 @@ You should have received a copy of the GNU Affero General Public License
 along with this program. If not, see <http://www.gnu.org/licenses/>.
 -->
 <template>
-  <div id="news-latest-view" class="px-2 pb-2">
-    <div class="article-container">
+  <div id="news-latest-view"  ref="news-latest-view" class="px-2 pb-2">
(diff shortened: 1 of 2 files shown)


webapp/src/main/webapp/news-list-view/components/views/NewsLatestView.vue
```diff
@@ -15,12 +15,12 @@ You should have received a copy of the GNU Affero General Public License
...
 <template>
-  <div id="news-latest-view" class="px-2 pb-2">
-    <div class="article-container">
+  <div id="news-latest-view"  ref="news-latest-view" class="px-2 pb-2">
+    <div :class="hasSmallWidthContainer ? 'article-small-container':'article-container'">
       <div
         v-for="(item, index) of newsInfo"
         :key="item"
-        class="article"
+        :class="hasSmallWidthContainer ? 'hasSmallWidthContainer' : 'article'"
```
ahamdi:
```suggestion
        :class="hasSmallWidthContainer ? 'smallWidthContainer' : 'article'"
```

webapp/src/main/webapp/skin/less/news.less
```diff
@@ -2903,6 +2908,151 @@
         transform: scale(1.06);
       }
     }
+    .hasSmallWidthContainer {
```
ahamdi:
```suggestion
    .smallWidthContainer {
```

ahamdi (approved):

PR #589, opened 2022-06-09
Task-57143: Adjust display of news headlines template using a small width container

hchaabene:


ahamdi (approved):
LGTM

PR #590, opened 2022-06-09
Task-57165: News mosaic template is ko if small width

hchaabene:
Prior to this change, first publish 4 articles with names not truncated , when i move latest new template in a small container or using mobile device to the right side or in a column , then choose mosaic model ,
This is due to changing the width is small, the display is bad and names of articles are truncated.
After this change, the article is readable , and names of articles are not truncated.


ahamdi (approved):
LGTM

PR #596, opened 2022-06-13
Task-58132: Sort news list of news application

sofyenne:
ISSUE : When user opens the draft list of application news and delete a last article in the list, Check the news application with the posted articles filter, the deleted article displayed on the top of posting articles list.
this issue is due getting list of news ordered by property `exo:dateModified`.
FIX : sort the list of news by creation date.


ahamdi (changes requested):
The list of News should be ordered by the modification date of the published content , not with the date from the original content.

PR #605, opened 2022-06-27
TASK-56899: Enhance latest news portlet and news illustration images load time

hakermi:
Prior to this change, a duplicated rest calls to retrieve the latest news each one took a lot of time to finish, a huge size of the retrieved data because of the retrieved binary contents of the illustration images and a lack of CacheControl use in illustrations rest endpoint.
This PR should :
- Fix the duplicated rest call for both latest news and news slider
- Avoid retrieving the binary contents of the illustration images
- Add CacheControl use while serving the news illustration images and use image/webp as mimeType for the cached images 
  content.

### DIFF
--- a/packaging/pom.xml
+++ b/packaging/pom.xml
@@ -24,7 +24,7 @@
   <parent>
     <artifactId>news</artifactId>
     <groupId>org.exoplatform.addons.news</groupId>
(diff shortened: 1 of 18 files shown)


mkrout (approved):
LGTM

PR #611, opened 2022-07-13
TASK-50471: Create a resized thumbnail image of news illustration when list news articles

hakermi:
Prior to this change, used illustrations in news list are displayed and retrieved with their original size which cause a perf issue and a latency while rendering these images by the browser.
This PR should return a resized thumbnail for news article illustrations for better performance and fast render by the browser


mkrout (dismissed):
LGTM

mkrout (approved):

PR #640, opened 2022-09-15
fix: Prevent display of alert template when there is no related articles - TASK-56232 - exoplatform/news

akhanfir:
Prior to this change, if there is no articles related to such template, the whole template's bloc is displayed. 
After this change, the Alerts block is no more displayed in this case.


azayati (approved):
LGTM

PR #646, opened 2022-10-21
fix: Fix missing number of reactions per period in contents analytics report - EXO-60102

IlhemEssaadi:
Prior to this change, the number of reactions per period in the content analysis report is not incremented when we like or comment on a post. This is caused by the event.getActivity() object passed to the NewsActivityListener whose templateParams is null and doesn't contain NEWS_ID param. After this fix, we retrieve the right stored news activity containing the needed template param based on the id of event.getActivity() which allows to broadcast the needed analytics events LIKE_NEWS and COMMENT_NEWS.


azayati (approved):
LGTM

PR #651, opened 2022-10-28
feat: Split news styles - EXO-60537

hakermi:
Prior to this change, news styles for all its portlets are added in one less page declared as a portal skin which is loaded in all pages which affect the stream page perf in our case by load unneeded styles. 
This PR splits news styles for each portlet and keep news extensions styles loaded in a dynamic container and presents in the stream page for the stream composer drawer and declare the news details styles as a portal skin for news activity details styles.


ahamdi:
Could you check the Sonar issues ?

hakermi:
> Could you check the Sonar issues ?

Done

ahamdi (approved):

PR #661, opened 2022-11-08
feat: News List template: use correct size for illustrations -EXO-60461

akhanfir:
Prior to this change, the size for illustrations in the templates lists news is still required at the original size of the image. After this change, the size of the illustrations is updated for the correct size.

### DIFF
--- a/webapp/src/main/webapp/news-list-view/components/views/NewsMosaicView.vue
+++ b/webapp/src/main/webapp/news-list-view/components/views/NewsMosaicView.vue
@@ -26,7 +26,10 @@ along with this program. If not, see <http://www.gnu.org/licenses/>.
           class="articleLink d-block"
           target="_self"
           :href="item.url">
-          <img :src="showArticleImage && item.illustrationURL !== null ? item.illustrationURL : '/news/images/news.png'" :alt="$t('news.latest.alt.articleImage')">
+          <img :src="showArticleImage && item.illustrationURL !== null ? 
+            $newsListService.illustrationURLResize(item.illustrationURL, `#top-news-mosaic #articleItem-${index}`)
+            : '/news/images/news.png'" 
+            :alt="$t('news.latest.alt.articleImage')">
           <div class="titleArea">
             <div v-if="showArticleDate" class="articleDate">
               <date-format
--- a/webapp/src/main/webapp/news-list-view/components/views/NewsSliderView.vue
+++ b/webapp/src/main/webapp/news-list-view/components/views/NewsSliderView.vue
@@ -31,7 +31,9 @@ along with this program. If not, see <http://www.gnu.org/licenses/>.
         dark>
         <v-img
           class="articleImage fill-height"
-          :src="showArticleImage && item.illustrationURL !== null ? item.illustrationURL : '/news/images/news.png'"
+          :src="showArticleImage && item.illustrationURL !== null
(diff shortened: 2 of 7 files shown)


webapp/src/main/webapp/news-list-view/components/views/NewsMosaicView.vue
```diff
@@ -147,6 +147,13 @@ export default {
         seeAllUrl: this.seeAllUrl,
       };
     },
+    illustrationURLResize(itemIllustrationURL) {
+      const element = document.querySelector('#top-news-mosaic .articleLink');
```
rdenarie: How it work if you have more than one .articlelink in #top-news-mosaic ?
I'm afraid that, for example, the second, third ... news take the same size as the first one.
Should we use the idItem ?

rdenarie (changes requested):
Is it possible to put function illustrationURLResize in a commun part ? to prevent code duplication ?

rdenarie (changes requested):
In addition, can you check sonar feedback ?

webapp/src/main/webapp/news-list-view/components/views/NewsMosaicView.vue
akhanfir: It’s true, I tested with only one news.

rdenarie (changes requested):
Is it possible to put function illustrationURLResize in a commun part ? to prevent code duplication ?

webapp/src/main/webapp/news-list-view/components/views/NewsSliderView.vue
```diff
@@ -165,6 +167,13 @@ export default {
         seeAllUrl: this.seeAllUrl,
       };
     },
+    illustrationURLResize(itemIllustrationURL) {
+      const element = document.querySelector('.sliderNewsItems');
+      if (element && itemIllustrationURL) {
+        return itemIllustrationURL.toString().concat('&size=')
+          .concat(element.clientWidth).concat('x').concat(element.clientHeight).toString();
+      }
+    }
```
rdenarie: to remove

webapp/src/main/webapp/news-list-view/js/NewsListService.js
```diff
@@ -51,4 +51,17 @@ export function saveSettings(saveSettingsURL, settings) {
       throw new Error('Response code indicates a server error', resp);
     }
   });
+}
+
+export function illustrationURLResize(illustrationURL, querySelector, idItem, newsListViewName) {
+  if ((idItem === 'articleItem-1' || idItem === 'articleItem-2' || idItem === 'articleItem-3')
+  && newsListViewName === 'newsLatest') {
+    return illustrationURL.concat('&size=107x107').toString();
+  } else {
```
rdenarie: Why, for this case, it is not possible to get element.height, and element.width ?

rdenarie (changes requested):

webapp/src/main/webapp/news-list-view/js/NewsListService.js
akhanfir: It is possible,but we can get it from the CSS file .

rdenarie: In fatc, I prefer to not have exceptions in the function, keep it simple, with only the if part.

rdenarie (changes requested):

rdenarie (approved):

PR #663, opened 2022-11-09
feat: update illustration size : EXO-60463 (#659)

GouadriaHanen:
prior to this change, In the news application, the illustration is requested in size 300x300 but displayed in a square 150x150 after this change, the size updated to use the correct one


rdenarie (approved):
LGTM

PR #670, opened 2022-11-17
 fix: Prefetch news list rest call contains the target hardcoded : EXO-60823

GouadriaHanen:
prior to this change, only the snapshotLatestNews are prefetched
after this change, all targets are prefetched

### DIFF
--- a/webapp/src/main/webapp/WEB-INF/jsp/newsListView.jsp
+++ b/webapp/src/main/webapp/WEB-INF/jsp/newsListView.jsp
@@ -69,8 +69,7 @@ along with this program. If not, see <http://www.gnu.org/licenses/>.
 
     PortalRequestContext rcontext = PortalRequestContext.getCurrentInstance();
     PortalHttpServletResponseWrapper responseWrapper = ( PortalHttpServletResponseWrapper ) rcontext.getResponse();
-    String target = viewTemplate.equals("NewsSlider") ? "snapshotSliderNews" : "snapshotLatestNews";
-    String newsListUrl = "/portal/rest/v1/news/byTarget/"+ target +"?offset=0&limit="+ limit + "&returnSize=true";
+    String newsListUrl = "/portal/rest/v1/news/byTarget/" + newsTarget + "?offset=0&limit=" + limit + "&returnSize=true";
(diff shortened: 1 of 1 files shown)


ahamdi (approved):

webapp/src/main/webapp/WEB-INF/jsp/newsListView.jsp
```diff
@@ -69,7 +69,8 @@ along with this program. If not, see <http://www.gnu.org/licenses/>.
 
     PortalRequestContext rcontext = PortalRequestContext.getCurrentInstance();
     PortalHttpServletResponseWrapper responseWrapper = ( PortalHttpServletResponseWrapper ) rcontext.getResponse();
-    String newsListUrl = "/portal/rest/v1/news/byTarget/snapshotLatestNews?offset=0&limit="+ limit + "&returnSize=true";
+    String target = viewTemplate.equals("NewsSlider") ? "snapshotSliderNews" : "snapshotLatestNews";
```
rdenarie: How it work for other view templates ? Mosaic ? Headlines ? Stories ?

Currenctly we have 2 opssible targets for news. But what if we add another one ? 
Can you check how the url is build in the newListApplication ? the target should be parameterized, then you should use the same parameter to build your prefetch url

rdenarie (changes requested):

webapp/src/main/webapp/WEB-INF/jsp/newsListView.jsp
GouadriaHanen: we have only two view sources latest  and slider news which can have a template as mosaic, stories headlines .. they are the template for displaying the information but they have the same source URL "snapshotSliderNews" or "snapshotLatestNews"

rdenarie: User can add a new target in target administration. For example, on devx instance, I add a new target "mytargetRomain". Then the url generated is https://devx-ft.exoplatform.org/portal/rest/v1/news/byTarget/mytargetRomain?offset=0&limit=4&returnSize=true
The target name is present in the url, and it is not snapshoSliderNews nor snapshotLatestNews

rdenarie (approved):

PR #671, opened 2022-11-18
feat: News List template: use correct size for illustrations -EXO-60461

akhanfir:
Prior to this change, When using:
- mosaic and headlines template, images are broken.
- news card and news list: news without images are not displayed.

After this change, the images are displayed correctly.


webapp/src/main/webapp/news-list-view/components/views/NewsMosaicView.vue
```diff
@@ -86,6 +86,7 @@ export default {
     this.reset();
     this.$root.$on('saved-news-settings', this.refreshNewsViews);
     this.getNewsList();
+    this.$newsListService.illustrationURLResize('test','test');
```
rdenarie: This should be removed ?

rdenarie (changes requested):

rdenarie (approved):

PR #676, opened 2022-11-24
fix: News Mosaic template : UI Improvements - EXO-60594

akhanfir:
After this change, we made some change in News Mosaic template UI side
1. Title should be extended to 5 lines long in the big left side article card
2. Little cards title should be extended to 2 lines long in the other 3 right side cards.
3. Header should be hidden if "show header" option is disabled
4. Setting button should be placed inside container not on header level.

### DIFF
--- a/webapp/src/main/webapp/news-list-view/components/NewsListView.vue
+++ b/webapp/src/main/webapp/news-list-view/components/NewsListView.vue
@@ -16,7 +16,7 @@ along with this program. If not, see <http://www.gnu.org/licenses/>.
 -->
 <template>
   <v-app class="news-list-view-app position-relative">
-    <v-card flat class="list-view-card rounded-0">
+    <v-card height="100%" flat class="list-view-card rounded-0">
       <v-card-text class="pa-0">
         <news-settings v-if="displayHeader" />
         <extension-registry-component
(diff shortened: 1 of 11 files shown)


webapp/src/main/webapp/news-list-view/components/NewsListView.vue
```diff
@@ -126,6 +129,9 @@ export default {
         componentOptions: this.selectedViewExtension,
       };
     },
+    isMobile() {
+      return this.$vuetify.breakpoint.width < 651;
```
ahamdi, line 133:
```suggestion
      return this.$vuetify.breakpoint.name === 'xs';
```

webapp/src/main/webapp/news-list-view/components/views/NewsMosaicView.vue
```diff
@@ -91,6 +96,9 @@ export default {
     isMobile() {
       return this.$vuetify.breakpoint.name === 'xs' || this.$vuetify.breakpoint.name === 'sm';
     },
+    isSmallBreakpoint() {
+      return this.$vuetify.breakpoint.width < 651;
```
ahamdi, line 100: we can not introduce a new viewport, we need to use existing ones . could you make needed changes ?

ahamdi (changes requested):

ahamdi (approved):

PR #679, opened 2022-11-29
fix: News List template: use correct size for illustrations -EXO-60461

akhanfir:
Prior to this change, the size for illustrations in the templates lists news is still required at the original size of the image. After this change, the size of the illustrations is updated for the correct size.


rdenarie (dismissed):
Please make the PR on develop and not in feature/maintenance

ahamdi (approved):

PR #682, opened 2022-11-29
fix : improve displaying newslist UI - EXO-60939

GouadriaHanen:
prior to this change, in the newsList template, there was too much empty space in the container, and the title is truncated in just one line. 
after this change, a UI improvement in displaying the newsList is added:
- Article titles should be truncated after  2 lines are reached. 
- edit padding to reduce unwanted white space


ahamdi:
@GouadriaHanen Could you add more details on what was done and why ?

GouadriaHanen:
> @GouadriaHanen Could you add more details on what was done and why?

done!

ahamdi (approved):

PR #691, opened 2022-12-07
feat: Edit target permissions from edit target drawer - EXO-60986

IlhemEssaadi:


services/src/main/java/org/exoplatform/news/service/impl/NewsTargetingServiceImpl.java
```diff
@@ -210,7 +210,7 @@ private NewsTargetingEntity toEntity(Metadata metadata) {
         NewsTargetingPermissionsEntity permissionEntity = new NewsTargetingPermissionsEntity();
         if (permission.contains("space")) {
           Space space = spaceService.getSpaceByPrettyName(List.of(permission.split(":")).get(1));
-          permissionEntity.setId(space.getId());
+          permissionEntity.setId("space:" + space.getDisplayName());
```
azayati: I think that this should be done in the previous PRs

webapp/src/main/webapp/news-publish-targets-management/components/NewsPublishTargetsManagementDrawer.vue
```diff
@@ -159,6 +159,7 @@ export default {
     originalTargetName: '',
     saveMode: 'creationMode',
     permissions: [],
+    permissionDisabled: true,
```
azayati: why do we need permissionDisabled ?

azayati (changes requested):
see review

webapp/src/main/webapp/news-publish-targets-management/components/NewsPublishTargetsManagementDrawer.vue
IlhemEssaadi: > why do we need permissionDisabled ?

To detect if there is any change for the list of permissions in case of edit to enable the button update.

services/src/main/java/org/exoplatform/news/service/impl/NewsTargetingServiceImpl.java
IlhemEssaadi: > I think that this should be done in the previous PRs

Done in the previous PRs !!

webapp/src/main/webapp/news-publish-targets-management/components/NewsPublishTargetsManagementDrawer.vue
azayati: IMHO it is should be better !permissionsUpdated

azayati (changes requested):

webapp/src/main/webapp/news-publish-targets-management/components/NewsPublishTargetsManagementDrawer.vue
IlhemEssaadi: > IMHO it is should be better !permissionsUpdated

Done !

azayati (approved):
LGTM

PR #700, opened 2022-12-12
feat: Edit target permissions from edit target drawer - EXO-60986

IlhemEssaadi:
Add a target permission section in edit target drawer hen editing a news target.


azayati (approved):
LGTM

PR #701, opened 2022-12-12
fix: images in shared news are not displayed - EXO-61176

GouadriaHanen:
before this change, images are not accessible from shared news of space I'm not a member of, since the user doesn't have the right to see them
after this change, the images are displayed successfully after adding the read permission for the shared news images


services/src/main/java/org/exoplatform/news/storage/jcr/JcrNewsStorage.java
```diff
@@ -1267,5 +1271,18 @@ private Session getSession(SessionProvider sessionProvider) throws RepositoryExc
                                                        .getDefaultWorkspaceName(),
                                       repositoryService.getCurrentRepository());
   }
-  
+
+  private  void updateNewsImagesPermissions(News news ,  SessionProvider sessionProvider, Space space  ) throws RepositoryException {
```
ahamdi:
```suggestion
  private  void updateNewsImagesPermissions(News news, SessionProvider sessionProvider, Space space) throws RepositoryException {
```

services/src/main/java/org/exoplatform/news/storage/jcr/JcrNewsStorage.java
```diff
@@ -1267,5 +1271,18 @@ private Session getSession(SessionProvider sessionProvider) throws RepositoryExc
...
+    Matcher matcher = IMAGE_SRC_PATTERN.matcher(news.getBody());
+    while (matcher.find()) {
+      String match = matcher.group(1);
+      String imageUUID = match.substring(match.lastIndexOf("/") + 1);
+      ExtendedNode image = (ExtendedNode) getNodeById(imageUUID, sessionProvider);
```
ahamdi, line 1280: Could you do a check here if the content is really present or not ?

services/src/main/java/org/exoplatform/news/storage/jcr/JcrNewsStorage.java
```diff
@@ -1267,5 +1271,18 @@ private Session getSession(SessionProvider sessionProvider) throws RepositoryExc
...
+      if (image.canAddMixin(EXO_PRIVILEGEABLE)) {
+        image.addMixin(EXO_PRIVILEGEABLE);
+      }
```
ahamdi: Is there a need to add this mixin ? you already loaded the image by UUID ?

services/src/test/java/org/exoplatform/news/storage/jcr/JcrNewsStorageTest.java
```diff
@@ -8,11 +8,8 @@
 import java.util.*;
 
 import javax.jcr.*;
-import javax.jcr.query.QueryManager;
```
ahamdi, line 11: Please add a new test for the new change ?

ahamdi (changes requested):

services/src/main/java/org/exoplatform/news/storage/jcr/JcrNewsStorage.java
GouadriaHanen: we could not set the permission without adding EXO_PRIVILEGEABLE

services/src/main/java/org/exoplatform/news/storage/jcr/JcrNewsStorage.java
GouadriaHanen: done

services/src/test/java/org/exoplatform/news/storage/jcr/JcrNewsStorageTest.java
GouadriaHanen: done

services/src/main/java/org/exoplatform/news/storage/jcr/JcrNewsStorage.java
ahamdi: Yes you are right, I misread it

ahamdi (approved):

PR #711, opened 2022-12-16
feat: Delete the space or group defined as a target permission should make it disallowed for any user - EXO-61359

IlhemEssaadi:


azayati (approved):
LGTM

PR #716, opened 2022-12-19
fix: news title UI - EXO-61363

GouadriaHanen:
prior to this change, the news title and summary are displayed with font size 14 px 
after this change,  the news title and summary are displayed with a font size 16 px

### DIFF
--- a/webapp/src/main/webapp/news-list-view/components/views/NewsLatestViewItem.vue
+++ b/webapp/src/main/webapp/news-list-view/components/views/NewsLatestViewItem.vue
@@ -30,9 +30,9 @@ along with this program. If not, see <http://www.gnu.org/licenses/>.
           class="spaceImage"
           :src="item.spaceAvatarUrl"
           :alt="$t('news.latest.alt.spaceImage')">
-        <span class="spaceName">{{ item.spaceDisplayName }}</span>
+        <span class="spaceName text-color text-body-1">{{ item.spaceDisplayName }}</span>
       </div>
-      <span v-if="showArticleTitle" class="articleTitle">{{ item.title }}</span>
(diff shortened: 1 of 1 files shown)


webapp/src/main/webapp/news-list-view/components/views/NewsListTemplateViewItem.vue
```diff
@@ -25,11 +25,11 @@ along with this program. If not, see <http://www.gnu.org/licenses/>.
     <div class="article-item-content">
       <span
         v-if="showArticleTitle"
-        class="text-color text-body-2"
+        class="text-body-1 black--text"
         :class="extraClass">{{ item.title }}</span>
       <span
         v-if="showArticleSummary"
-        class="text-color text-body-2"
+        class="text-body-1 black--text"
```
ahamdi: shouldn't we need to use primary text color, not black one ?

ahamdi (changes requested):

webapp/src/main/webapp/news-list-view/components/views/NewsListTemplateViewItem.vue
GouadriaHanen: in the task " article title style must be the same as displayed documents titles in Documents widget" which is in black

ahamdi: I think we need to use this color https://github.com/Meeds-io/platform-ui/blob/develop/platform-ui-skin/src/main/webapp/skin/less/helpers.less#L40
could you check with Ikbel ?

GouadriaHanen: done!

ahamdi (approved):

PR #698, opened 2022-12-12
feat: Display news targets according to the permission in publish section of drawers - EXO-60989

IlhemEssaadi:

PR #721, opened 2022-12-21
feat: Display allowed news targets in publish section of publishing drawers - EXO-60989 (#698)

IlhemEssaadi:
We will display the allowed news targets in the publish section of the different publishing drawers.


azayati (approved):
LGTM

PR #728, opened 2022-12-23
feat: Display No target is allowed in news list settings - EXO-61246

IlhemEssaadi:


webapp/src/main/webapp/news-list-view/components/settings/NewsSettingsDrawer.vue
```diff
@@ -196,6 +197,9 @@ export default {
     saveSettingsURL: ''
   }),
   computed: {
+    backgroundColor(){
+      return this.newsTargets.length === 0 ? '#E1E8EE': '';
```
azayati, line 201: can we use a css class or a color variable instead ?

azayati (changes requested):

azayati (approved):
LGTM

PR #735, opened 2022-12-27
Fix: image isn't considered as a content appended to the news body so the post button is disabled - EXO-60812

hthabti:
prior to this change, when a user append an image in the body of the news during the creation or modification process, the post button is disabled.
after this change the post button is enabled when the appended content in the body is an image


ahamdi (changes requested):
When the image is inserted, the autoSave is fired in Create/Update article. I think that whenever we have a saved Draft, the Post button should be activated.
Could you change the logic of Post button, it does not have to rely on the content as it is HTML and could be escaped when transformed to article.

hthabti:
@ahamdi I put back the line of the getcontent() as it was + made some edits on the if condition for postDisabled(). Thus saving drafts won't be affected and the button would be enabled when the content is an image

exo-swf:

exo-swf:

webapp/src/main/webapp/news-activity-composer-app/components/ExoNewsActivityComposer.vue
```diff
@@ -394,17 +394,24 @@ export default {
...
+      }
     },
     updateDisabled: function () {
       // disable update button while uploading an attachment
       if (this.uploading) {
         return true;
       }
       // disable update button if a mandatory field is empty
-      if (!this.news.title || !this.news.title.trim() || !this.news.body || !this.news.body.replace(/&nbsp;/g, '').trim()) {
+      if (!this.news.title || !this.news.title.trim() || !this.news.body.replace(/&nbsp;/g, '').trim() || new DOMParser().parseFromString(this.news.body, 'text/html').documentElement.querySelectorAll('img').length === new DOMParser().parseFromString(this.originalNews.body, 'text/html').documentElement.querySelectorAll('img').length && this.news.body === this.originalNews.body) {
```
ahamdi: we can not rely on the number of images, as we can add and remove an image in the same time and it won't be considered as a modification of an article

ahamdi (changes requested):

webapp/src/main/webapp/news-activity-composer-app/components/ExoNewsActivityComposer.vue
hthabti: change done as follows
`      if (!this.news.title || !this.news.title.trim() || !this.news.body || !this.news.body.replace(/&nbsp;/g, '').trim()) {
        return true;
      }
      } else if (this.news.body.replace(/&nbsp;|<div>|<\/div>|<p>|<\/p>|\s+/g, '') !== this.originalNews.body.replace(/&nbsp;|<div>|<\/div>|<p>|<\/p>|\s+/g, '').replace(/\/>/g, '>').replace(/&#61;/g, '=')) {
        return false;
      } `

hbenali:
Stale PR

PR #749, opened 2023-01-05
fix: Unpublish news problem after editing news title - EXO-61538

GouadriaHanen:
before this change, after changing the title of the published news, the Unpublish of the news can't be done and the publish button keeps loading, since it can't find the shared news with its new title, it throws an exception that is not handled
after this change, after changing the news title, the published news is removed and the updated news is published


ahamdi (commented):
This definitely deserves a Unit test

ahamdi (approved):

PR #753, opened 2023-01-06
feat: Disable publish checkbox, hide targets list and audiencing in different publishing drawers when no target is allowed - EXO-61512

IlhemEssaadi:


azayati (approved):
LGTM

PR #778, opened 2023-01-13
feat: Fix blank displaying of news with Only space member audience for a non space member - EXO-61721 

IlhemEssaadi:


azayati (approved):
LGTM

PR #788, opened 2023-01-17
feat: Fix wrong default audience value and default targets when edit publishing and scheduling - EXO-61777

IlhemEssaadi:
Prior to this change:

- When we unpublish a published article then publish it again, the displayed audience value is "only space members"
-  When we unpublish a scheduled article then publish it again, the displayed audience value is "only space members" and the old targets still displayed

After this commit, after unpublishing, we will ensure to reset the audience value to "all users" for published and scheduled articles, and reset the targets with empty values for scheduled articles.


azayati (approved):
LGTM

PR #809, opened 2023-01-31
fix: Standardize font styles/size between notes and news pages - EXO-61667.

akhanfir:
Prior to this change, font styles and size were not standardized between different views in notes and news. After this change, some css style are applied on headline ,bulleted-list, numbered-list and block quote to be standardize with notes pages.


ahamdi (commented):
Could you avoid duplicating the rules ?

ahamdi (approved):

PR #815, opened 2023-02-15
fix: publish news notification is received when editing a published news - EXO-62469

GouadriaHanen:
prior to this change, when editing published news, notification of publishing news is received since it is republished 
after this change, avoid sending a new notification if the news is already published and no more create symlink for published news

### DIFF
--- a/services/src/main/java/org/exoplatform/news/service/NewsService.java
+++ b/services/src/main/java/org/exoplatform/news/service/NewsService.java
@@ -75,9 +75,10 @@ public interface NewsService {
    *
    * @param news to be published
    * @param publisher of the News
+   * @param broadcast if to broadcast send notification
    * @throws Exception when error
    */
-  void publishNews(News news, String publisher) throws Exception;
+  void publishNews(News news, String publisher, boolean broadcast) throws Exception;
   
   /**
    * Unpublish a News
--- a/services/src/main/java/org/exoplatform/news/service/NewsService.java
+++ b/services/src/main/java/org/exoplatform/news/service/NewsService.java
@@ -75,10 +75,9 @@ public interface NewsService {
    *
    * @param news to be published
    * @param publisher of the News
-   * @param broadcast if to broadcast send notification
    * @throws Exception when error
    */
-  void publishNews(News news, String publisher, boolean broadcast) throws Exception;
+  void publishNews(News news, String publisher) throws Exception;
   
   /**
    * Unpublish a News
--- a/services/src/main/java/org/exoplatform/news/service/impl/NewsServiceImpl.java
+++ b/services/src/main/java/org/exoplatform/news/service/impl/NewsServiceImpl.java
@@ -158,29 +158,14 @@ public class NewsServiceImpl implements NewsService {
     org.exoplatform.services.security.Identity updaterIdentity = NewsUtils.getUserIdentity(updater);
     boolean canPublish = NewsUtils.canPublishNews(news.getSpaceId(), updaterIdentity);
     Set<String> previousMentions = NewsUtils.processMentions(originalNews.getOriginalBody());
-    try {
-      // edit title case
-      if (StringUtils.isNotBlank(news.getTitle()) && !news.getTitle().equals(originalNews.getTitle())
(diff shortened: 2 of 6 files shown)


services/src/main/java/org/exoplatform/news/service/NewsService.java
```diff
@@ -75,9 +75,10 @@ public interface NewsService {
    *
    * @param news to be published
    * @param publisher of the News
+   * @param broadcast if to broadcast listener
    * @throws Exception when error
    */
-  void publishNews(News news, String publisher) throws Exception;
+  void publishNews(News news, String publisher, boolean broadcast) throws Exception;
```
ahamdi: We should not change the API in a maintenance task, could you add another function with the new parameter ?

services/src/test/java/org/exoplatform/news/service/impl/NewsServiceImplTest.java
```diff
@@ -170,22 +170,18 @@ public void testUpdateNews() throws Exception {
         news.setPublished(true);
         news.setPublicationState("archived");
         news.setCanPublish(true);
-        newsService.updateNews(news, "root", null, true);
-        verify(newsStorage, times(0)).unpublishNews(news.getId());
-        verify(newsStorage, times(1)).publishNews(any(News.class));
-
```
ahamdi: Why do you need to remove this check ?

ahamdi (changes requested):

services/src/main/java/org/exoplatform/news/service/NewsService.java
GouadriaHanen: done!

services/src/test/java/org/exoplatform/news/service/impl/NewsServiceImplTest.java
GouadriaHanen: restored

services/src/main/java/org/exoplatform/news/service/impl/NewsServiceImpl.java
```diff
@@ -166,12 +166,12 @@ public News updateNews(News news, String updater, Boolean post, boolean publish)
         if (news.getTargets() == null || news.getTargets().isEmpty()) {
           news.setTargets(oldTargets);
         }
-        publishNews(news, updater);
+        publish(news, updater);
```
azayati: As discussed together last time, we should instead of that just restore the old code and nomore create symlink when publishing an article which will avoid [the last fixed problem](https://github.com/exoplatform/news/commit/f8c35f2fcc58bf37cb93ae0717d680acfb0fb763). So IMO [this commit ](https://github.com/exoplatform/news/commit/f8c35f2fcc58bf37cb93ae0717d680acfb0fb763) should be reviewed.
@GouadriaHanen @ahamdi

azayati (changes requested):

services/src/main/java/org/exoplatform/news/storage/jcr/JcrNewsStorage.java
```diff
@@ -634,6 +609,8 @@ public News updateNews(News news, String updater) throws Exception {
       newsNode.setProperty("exo:body", news.getBody());
       newsNode.setProperty("exo:dateModified", Calendar.getInstance());
       newsNode.setProperty(EXO_NEWS_LAST_MODIFIER, updater);
+      newsNode.setProperty("exo:pinned", news.isPublished());
```
azayati: Normally this method is called even when we update the simple fields (title, summary, content) without updading publish, so IMO we should not set the property exo:pinned in that case which may causes a problem (to be verified)
I think that we should keep publishNews and unpublishNews methods in jcr storage layer with only setting exo:pinned property (without creating symlink)
cc @ahamdi

azayati (dismissed):
@GouadriaHanen we should also think about an upgrade plugin which deletes the folder /Application Data/News/Pinned 
cc @ahamdi

ahamdi (dismissed):

ahamdi (approved):

PR #847, opened 2023-04-25
Feat: display views count on news list template -EXO-63953

hthabti:
prior to this change only likes and comments count are displayed for news list template.
after this change views count is displayed in addition to likes & comments


webapp/src/main/webapp/news-list-view/components/views/NewsListTemplateViewItem.vue
```diff
@@ -58,6 +58,10 @@ along with this program. If not, see <http://www.gnu.org/licenses/>.
             mdi-comment
           </v-icon>
           <div class="comments-count me-2">{{ item.commentsCount }}</div>
+          <v-icon class="reactionIconStyle me-1" size="12">
```
ahamdi: shouldn't it be `reactions-icon-style` instead of `reactionIconStyle` ?

ahamdi (approved):

PR #850, opened 2023-05-03
fix: The search for news using the news tag text does not display any results - EXO-64009

sofyenne:
Before to this change on news app when we searched for news by typing the text of the tag, no results were displayed. The issue was that the news query was not built to include the filter by the tag text.

This change will find the tag names by the search text to assume that the searched text is a tag text and  update the build query method to enable filtering by the tag text.

### DIFF
--- a/services/src/main/java/org/exoplatform/news/connector/NewsSearchConnector.java
+++ b/services/src/main/java/org/exoplatform/news/connector/NewsSearchConnector.java
@@ -1,6 +1,7 @@
 package org.exoplatform.news.connector;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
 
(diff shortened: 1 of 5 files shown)


services/src/main/java/org/exoplatform/news/connector/NewsSearchConnector.java
```diff
@@ -66,6 +67,10 @@ public List<SearchResult> search(NewsFilter filter,
             repositoryService.getCurrentRepository());
 
     if (filter.getSearchText() != null) {
+      // search text composed by one keyword maybe it is the tag text
+      if (filter.getSearchText().split(" ").length <= 1){
```
ahamdi: We need to be sure if the text is a tag or a simple text, we can not guess it based on the number of words in the query

ahamdi (changes requested):

services/src/main/java/org/exoplatform/news/connector/NewsSearchConnector.java
sofyenne: done ! Using tag service we will find the tag names based on the search text to ensure that the search text is a tag text .

ahamdi (approved):

PR #852, opened 2023-05-08
fix: The search for news using the news tag text does not display any results - EXO-64009 (#850)

sofyenne:
Before to this change on news app when we searched for news by typing the text of the tag, no results were displayed. The issue was that the news query was not built to include the filter by the tag text.

This change will find the tag names by the search text to assume that the searched text is a tag text and update the build query method to enable filtering by the tag text.


exo-swf:

Jihed525 (approved):

PR #864, opened 2023-06-08
fix: [NEWS] Gif without animation in news illustration EXO-64163.

akhanfir:
Prior to this change, when create news in space whoose illustration image is a gif and post it, the gif is without animation. To fix this problem, when the image is of type gif, change the size to null to cancel converting it to an image. After this change, illustration image is animated as it is the case in edit mode.

### DIFF
--- a/services/src/main/java/org/exoplatform/news/rest/NewsRestResourcesV1.java
+++ b/services/src/main/java/org/exoplatform/news/rest/NewsRestResourcesV1.java
@@ -783,7 +783,8 @@ public class NewsRestResourcesV1 implements ResourceContainer, Startable {
           String[] dimension = size.split("x");
           byte[] thumbnail = thumbnailService.createCustomThumbnail(news.getIllustration(),
                                                                     Integer.parseInt(dimension[0]),
-                                                                    Integer.parseInt(dimension[1]));
+                                                                    Integer.parseInt(dimension[1]),
+                                                                    news.getIllustrationMimeType());
           news.setIllustration(thumbnail);
         }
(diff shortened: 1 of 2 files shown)


services/src/main/java/org/exoplatform/news/rest/NewsRestResourcesV1.java
```diff
@@ -776,6 +776,9 @@ public Response getNewsIllustration(@Context Request rsRequest,
       EntityTag eTag = (size == null || size.isBlank()) ? new EntityTag(String.valueOf(lastUpdated)) : new EntityTag(lastUpdated+"-"+size);
       Response.ResponseBuilder builder = rsRequest.evaluatePreconditions(eTag);
       if (builder == null) {
+        if (news.getIllustrationMimeType().equalsIgnoreCase("image/gif")) {
```
ahamdi, line 779: Just a question, is the image cached correctly by the browser ?

ahamdi:
```suggestion
        if ("image/gif".equalsIgnoreCase(news.getIllustrationMimeType())) {
```
Just a question, is the image cached correctly by the browser ?

rdenarie: By doing this in NewsRestService, we only make the work for news.
We should do the modification in ResizeService, so that all image resized in the platform will benefit of the fix (new, activities, avatar, banner ...)

rdenarie (dismissed):

ahamdi (approved):

ahamdi (approved):

PR #895, opened 2023-07-27
 fix: [NEWS] Filtered articles aren't retreived if they contain links - EXO-64780

akhanfir:
Prior to this change, when create an article with a link, open the news app, and filter articles by the keyword of the link, that article is not picked up in the filtered list. To fix this problem, add in the SQL query the search for this term in the property `exo:body`. After this change, this article is listed in the results.


services/src/main/java/org/exoplatform/news/queryBuilder/NewsQueryBuilder.java
```diff
@@ -54,7 +54,9 @@ public StringBuilder buildQuery(NewsFilter filter) throws Exception {
         }
         if (filter.getSearchText() != null && !filter.getSearchText().equals("")) {
           String escapedQuoteSearchText = filter.getSearchText().replace("'", "''").replace("\"", "\"\"");
-          sqlQuery.append("CONTAINS(.,'").append(escapedQuoteSearchText).append("')");
+          int index = escapedQuoteSearchText.indexOf("~0.6");
```
ahamdi: The fuzzy index should never be used to extract the searched word as the index may be changed.

services/src/main/java/org/exoplatform/news/queryBuilder/NewsQueryBuilder.java
```diff
@@ -54,7 +54,9 @@ public StringBuilder buildQuery(NewsFilter filter) throws Exception {
...
+          String searchTextTerm = escapedQuoteSearchText.substring(0, index);
+          sqlQuery.append("(CONTAINS(.,'").append(escapedQuoteSearchText).append("') OR (exo:body LIKE '%").append(searchTextTerm).append("%'))");
```
ahamdi: the directive Contains will look for the exact word, when the word is put inside a HTML tag with no spaces before and after it won't be returned. 
You can remove the usage of CONTAINS but you have to include other properties of News in the search query exo:summary, exo:title.
A better solution would be to use Elastic search instead of searching the JCR. Could you evaluate the complexity of such change ?

ahamdi (changes requested):

services/src/main/java/org/exoplatform/news/queryBuilder/NewsQueryBuilder.java
ahamdi: @akhanfir my comment here is still relevent. where is this index added ? we should have the exact text without the fuy index

ahamdi (commented):
@akhanfir my comment here is still relevent. where is this index added ? we should have the exact text without the fuy index

services/src/main/java/org/exoplatform/news/queryBuilder/NewsQueryBuilder.java
```diff
@@ -54,7 +55,9 @@ public StringBuilder buildQuery(NewsFilter filter) throws Exception {
         }
         if (filter.getSearchText() != null && !filter.getSearchText().equals("")) {
           String escapedQuoteSearchText = filter.getSearchText().replace("'", "''").replace("\"", "\"\"");
-          sqlQuery.append("CONTAINS(.,'").append(escapedQuoteSearchText).append("')");
+          int index = escapedQuoteSearchText.indexOf(NewsSearchConnector.FUZZY_SEARCH_SYNTAX);
```
ahamdi: Could you add the fuzzy search index as a parameter to the object filter

exo-swf:

services/src/main/java/org/exoplatform/news/filter/NewsFilter.java
```diff
@@ -19,6 +19,8 @@ public class NewsFilter {
 
   private String       searchText;
 
+  private String       originText;
```
ahamdi: Could you add the fuzzy index here instead of duplicating the original searched text ? this fuzzy index could b e assigned to the filter inside the class https://github.com/exoplatform/news/blob/e3b16f21e2377b1b58a165211e5a720afdd3d8a1/services/src/main/java/org/exoplatform/news/connector/NewsSearchConnector.java#L69

ahamdi (changes requested):

ahamdi (approved):

PR #904, opened 2023-08-08
 fix:[news] '#' glitches the filter - EXO-65188

akhanfir:
Prior to this change, when create a tag in article and go to news app make a search for that tag(# followed by tag name), there is no filter applied, the display remain the same. After this change, return results with the searched tag.

### DIFF
--- a/services/src/main/java/org/exoplatform/news/rest/NewsRestResourcesV1.java
+++ b/services/src/main/java/org/exoplatform/news/rest/NewsRestResourcesV1.java
@@ -483,8 +483,12 @@ public class NewsRestResourcesV1 implements ResourceContainer, Startable {
         String lang = request.getLocale().getLanguage();
         TagService tagService = CommonsUtils.getService(TagService.class);
         long userIdentityId = RestUtils.getCurrentUserIdentityId();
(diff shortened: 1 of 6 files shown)


exo-swf:

services/src/main/java/org/exoplatform/news/rest/NewsRestResourcesV1.java
```diff
@@ -483,8 +483,12 @@ public Response getNews(@Context HttpServletRequest request,
         String lang = request.getLocale().getLanguage();
         TagService tagService = CommonsUtils.getService(TagService.class);
         long userIdentityId = RestUtils.getCurrentUserIdentityId();
-        List<TagName> tagNames = tagService.findTags(new TagFilter(text, 0), userIdentityId);
-        if (tagNames != null && !tagNames.isEmpty()) newsFilter.setTagNames(tagNames.stream().map(e -> e.getName()).toList());
+        if (text.indexOf("#") == 0) {
```
ahamdi, line 486: Does this work for cases where we have a mix of normal text and tags ?

ahamdi (approved):

PR #905, opened 2023-08-09
fix :  Inserted images/videos aren't considred as a content in news - EXO-65311

sofyenne:
Before this change, when a user inserted an image or a video into the body of the news during the creation or modification process, the post button would be disabled. After this change, the post button is enabled when the  news body isn't empty and the update button is enabled when a new content is inserted.

### DIFF
--- a/webapp/src/main/webapp/news-activity-composer-app/components/ExoNewsActivityComposer.vue
+++ b/webapp/src/main/webapp/news-activity-composer-app/components/ExoNewsActivityComposer.vue
@@ -396,19 +396,31 @@ export default {
       return this.activityId && this.activityId !== '';
     },
     isEmptyNewsBody() {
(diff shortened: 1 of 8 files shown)


webapp/src/main/webapp/news-activity-composer-app/components/ExoNewsActivityComposer.vue
```diff
@@ -395,23 +395,29 @@ export default {
     editMode: function() {
       return this.activityId && this.activityId !== '';
     },
+    hasMediaContent() {
+      const parsedDocument = new DOMParser().parseFromString(this.news.body, 'text/html');
+      const mediaElements = parsedDocument.querySelectorAll('img, video, iframe');
```
ahamdi: Unfortunately this is still insufficient as it is limited to those tags and does not cover all possible HTML tags that could be added (svg, canva, audio, etc ...) even if it not the case now, but everytime a new tag is allowed, we will have to update the fix.
I prefer doing comparison of raw HTML content

exo-swf:

ahamdi (approved):

PR #910, opened 2023-08-11
feat: Enhance news load resources using load-group - EXO-63805

hakermi:
Prior to this change, all news extensions were loaded inside all-bottom-container including the non needed once in all pages. This PR add the use of load-group mechanism to enhance the resources load and load only what is needed in the current page.


rdenarie (changes requested):
Can you check the sonar feedback marked as BUG ?

services/src/main/java/org/exoplatform/news/rest/NewsRestResourcesV1.java
```diff
@@ -580,6 +580,13 @@ public Response getNewsByActivityId(@Parameter(description = "Activity id", requ
         return Response.status(Response.Status.NOT_FOUND).build();
       }
       news.setIllustration(null);
+      Identity userIdentity = identityManager.getOrCreateUserIdentity(currentIdentity.getUserId());
+      if (userIdentity != null) {
+        news.setFavorite(favoriteService.isFavorite(new Favorite("news",
+                                                                 news.getId(),
+                                                                 "",
+                                                                 Long.parseLong(userIdentity.getId()))));
+      }
```
rdenarie, line 589: This should be removed ?

webapp/src/main/webapp/WEB-INF/conf/news/nodetypes-templates/news/views/view1.gtmpl
```diff
@@ -152,6 +152,7 @@
         activityId: '${activityId}',
         hiddenSpace: ${news.hiddenSpace},
         isSpaceMember: ${news.isSpaceMember},
+        favorite: ${news.isFavorite}
```
rdenarie, line 155: Why do you need this variable ?

exo-swf:

hakermi:
> Can you check the sonar feedback marked as BUG ?

Done

services/src/main/java/org/exoplatform/news/rest/NewsRestResourcesV1.java
hakermi: why it should be removed? i have faced an existing old issue and i fixed it.

webapp/src/main/webapp/WEB-INF/conf/news/nodetypes-templates/news/views/view1.gtmpl
hakermi: https://github.com/exoplatform/news/pull/910#discussion_r1297132987

services/src/main/java/org/exoplatform/news/rest/NewsRestResourcesV1.java
hakermi: the issue is that when access news details and that you have set the news as favorite, you will still have the option to set it again and the icon of favorite is empty icon which indicates that news is not marked as favorite while it's the case

rdenarie (approved):
ok for me

PR #918, opened 2023-08-18
feat: Enhance news load resources using load-group - EXO-63805

hakermi:
Prior to this change, all news extensions were loaded inside all-bottom-container including the non needed once in all pages. 
This PR add the use of load-group mechanism to enhance the resources load and load only what is needed in the current page.


rdenarie (approved):
PR approved, but can you update the commit message when merging ? It is a fix and not a feature.

Thanks

hakermi:
> PR approved, but can you update the commit message when merging ? It is a fix and not a feature.
> 
> Thanks

Done

PR #921, opened 2023-08-25
feat: Add  images portlet to Content category - EXO-65944 - Meeds-io/MIPs#68

IlhemEssaadi:


webapp/src/main/webapp/WEB-INF/conf/news/application-registry-configuration.xml
```diff
@@ -76,6 +76,32 @@
                     </field>
                   </object>
                 </value>
+                <value>
```
azayati, line 79: this configuration should not be added here in news codebase but into layout-management codebase, the same location of the portlet itself. This PR should be closed

azayati (changes requested):

PR #950, opened 2023-10-19
fix: Article drafts edition makes the article displayed on top of the news app - EXO-56012

akhanfir:
Before this fix, when post several articles in a space and in news app, click on three dots>edit button for the last posted article then type some inputs then click on cancel, this article is displayed on the top of the posted articles list. in fact, it is that the draft which was created and the article existing in the same property of the date modified (dateModified) to fix this problem, during creation of draft in set another property (lastModifiedDate) and during get all the articles by modifying the tree property according to the filter of this list if it is drafted by using it as a property lastModifiedDate if not by using the same existing property. After this change, the routed list is sorted either by lastModifiedDate if the filter is the draft article list or by dateModified if the list has another filter.

### DIFF
--- a/services/src/main/java/org/exoplatform/news/rest/NewsRestResourcesV1.java
+++ b/services/src/main/java/org/exoplatform/news/rest/NewsRestResourcesV1.java
@@ -1017,13 +1017,16 @@ public class NewsRestResourcesV1 implements ResourceContainer, Startable {
           break;
         }
       }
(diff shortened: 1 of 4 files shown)


ahamdi (changes requested):
Could you add a Unit test for this fix ?

ahamdi (approved):

PR #956, opened 2023-11-01
fix: Article drafts edition makes the article displayed on top of the news app - EXO-56012

akhanfir:
Before this fix, when post several articles in a space and in news app, click on three dots>edit button for the last posted article then type some inputs then click on cancel, this article is displayed on the top of the posted articles list. in fact, it is that the draft which was created and the article existing in the same property of the date modified (dateModified) to fix this problem, during creation of article in set another property (publication:liveDate) and during get all the articles by modifying the tree property according to the filter of this list if it is drafted by using it as a property (exo:dateModified) otherwise using the previous property. After this change, the routed list is sorted either by (exo:dateModified) if the filter is the draft article list or by (publication:liveDate) if the list has another filter.

### DIFF
--- a/services/src/main/java/org/exoplatform/news/storage/jcr/JcrNewsStorage.java
+++ b/services/src/main/java/org/exoplatform/news/storage/jcr/JcrNewsStorage.java
@@ -890,13 +890,17 @@ public class JcrNewsStorage implements NewsStorage {
    * @throws RepositoryException when error
    */
   private Date getPublicationDate(Node node) throws RepositoryException {
-    VersionNode versionNode = new VersionNode(node, node.getSession());
-    List<VersionNode> versions = versionNode.getChildren();
-    if (!versions.isEmpty()) {
-      versions.sort(Comparator.comparingInt(v -> Integer.parseInt(v.getName())));
-      return versions.get(0).getCreatedTime().getTime();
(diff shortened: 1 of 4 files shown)


services/src/main/java/org/exoplatform/news/storage/jcr/JcrNewsStorage.java
```diff
@@ -890,13 +890,17 @@ public String substituteUsernames(String portalOwner, String message) {
    * @throws RepositoryException when error
    */
   private Date getPublicationDate(Node node) throws RepositoryException {
-    VersionNode versionNode = new VersionNode(node, node.getSession());
-    List<VersionNode> versions = versionNode.getChildren();
-    if (!versions.isEmpty()) {
-      versions.sort(Comparator.comparingInt(v -> Integer.parseInt(v.getName())));
-      return versions.get(0).getCreatedTime().getTime();
+    if (node.hasProperty("publication:liveDate") == true) {
```
rdenarie:
```suggestion
    if (node.hasProperty("publication:liveDate")) {
```

rdenarie (dismissed):

exo-swf:

akhanfir:
I have changed that you ask me, problem is fixed. @rdenarie

ahamdi (approved):

PR #977, opened 2023-11-13
fix: button favorite and 3 dots menu in activity detail have no accessible name - EXO-62416 .

akhanfir:
Before this change, after creating an activity in a space when you open DevTools of your browser and in the flagship tab, run the accessibility check, the favorite button and the 3 dots menu button have no accessible name. To resolve this problem, in the News_eu.properties file, added the labels following find it easily when searching, Remove from favorites and Open activity menu to the Archive article, Unarchive article and menu 3 buttons points of the activity. After this change, these faulty elements are no longer displayed in the flagship tab, launch accessibility.


services/src/main/resources/locale/portlet/news/News_en.properties
```diff
@@ -88,15 +88,15 @@ news.activity.clickToShowDetail=Click here for more news details
 
 news.archive.btn.confirm=Confirm
 news.archive.btn.cancel=Cancel
-news.archive.action=Archive article
+news.archive.action=Bookmark it to easily find it when searching
```
rdenarie: Why this value is changed ?

services/src/main/resources/locale/portlet/news/News_en.properties
```diff
@@ -88,15 +88,15 @@ news.activity.clickToShowDetail=Click here for more news details
...
 news.archive.confirm=This article will be archived and will no longer be available for users. You still can unarchive it from News application.
 news.archive.success=The article was successfully archived
 news.archive.error=An error occurred when trying to archive this article. Try again, if the error persists, please contact your administrator.
 news.archive.label=Archived
 
 news.archive.text=Sorry, you don't have access to this article. Please contact your administrator.
 
-news.unarchive.action=Unarchive article
+news.unarchive.action=Remove from favorites
```
rdenarie: Why this value is changed ?

rdenarie (dismissed):

services/src/main/resources/locale/portlet/news/News_en.properties
akhanfir: This was requested in the task, but if you wish, we can keep the old text.

rdenarie: can you confirm that the key news.archive.action correspond to "Favorite" action ? 
It seems related to different things

rdenarie: In addition, labels like that must be changed in crowdin

services/src/main/resources/locale/portlet/news/News_en.properties
akhanfir: As you said preferably that we add another key to avoid changing another label so it must that I add another key and I would change the description of the PR.

services/src/main/resources/locale/portlet/news/News_en.properties
akhanfir: As you said preferably that we add another key to avoid changing another label so it must that I add another key and I would change the description of the PR.

ahamdi (approved):

PR #999, opened 2023-11-28
fix: User and space avatar are not accessible - EXO-62527

akhanfir:
Before this change, when create an activity and refresh stream, the a tag of avatar have not a discernible name, the img tag for avatar have a role=presentation which is not the correct one and space and user avatar are concerned. To fix this, added:
  -the arial-label for this a tag of user avatar by the key news.avatar.author.title which returns a text Open {userFullname} profile.
  -the alt for this image tag by the key news.avatar.author.alt which returns a text{userFullname}'s avatar.
  -the arial-label for this a tag of space avatar by the key news.space.icon.title which returns a text Open {userFullname} profile.
  -the alt for this image tag by the key news.space.icon.alt which returns a text{userFullname}'s avatar.
After this change, user and space avatar are accessible.

### DIFF
--- a/services/src/main/resources/locale/portlet/news/News_en.properties
+++ b/services/src/main/resources/locale/portlet/news/News_en.properties
@@ -162,6 +162,12 @@ news.cards.readMore=Read more
 news.details.archive.action=Bookmark it to easily find it when searching
 news.details.unarchive.action=Remove from favorites
 news.details.menu.open=Open activity menu
+
+news.space.icon.title=Open space {0}
+news.space.icon.alt=Space {0} avatar
+news.avatar.author.title=Open {0} profile
+news.avatar.author.alt={0}'s avatar
+
 news.details.header.menu.share=Share
 news.details.header.menu.edit=Edit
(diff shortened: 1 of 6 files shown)


rdenarie (commented):
Does this modification can be done in ExoAvatar directly ?
So that it is applied everywhere the component is used.
With your modification, it is only done in news

akhanfir:
In this case where I made the changes we use the v-avatar tag not exo-avatar

Jihed525:
@rdenarie can you check this PR ?

rdenarie:
Can we rebase the pr ?

Jihed525:
> Can we rebase the pr ?

should be ok now, can you check @rdenarie ?

rdenarie:
As I understand, in news list, we do not use exo-avatar or space-avatar.

You already fixed both in social, but to fix in news list, you need to do it here, is it right ?

cc @akhanfir

akhanfir:
yes, exactly that @rdenarie

rdenarie (approved):

PR #1042, opened 2023-12-22
fix: Fix Image Alignment Issue After Saving - EXO-68298

IlhemEssaadi:
Prior to this change, the image was correctly aligned by the middle in height with the text while in edit mode, but after saving, it would align by the bottom with the text.This commit ensures that the image maintains a consistent vertical alignment with the text after saving.


azayati (approved):
LGTM

PR #1049, opened 2023-12-28
fix: Fix news settings form grayed out and not editable - EXO-68547

IlhemEssaadi:
Prior to this change, the News List View portlet's settings drawer appeared in gray and was non-editable. This commit fixes the problem by implementing event listeners to dynamically switch the position style of the #stickyBlockDesktop element between "sticky" and "static" based on drawer events.


azayati (approved):
LGTM

PR #1052, opened 2024-01-02
fix : Fix user mentions in news article - EXO-68586

sofyenne:
Prior to this change the user mentions did not worked in the news article , this change is going to add the suggester plugin to the CKEditor,  resolving this issue .


azayati (approved):
LGTM

PR #1062, opened 2024-01-18
feat: Group Mentioning meeds-io/MIPs#109

rdenarie:
This commit adapt code for new feature Group Mentioning : it change the News-specific function which substiture mention to the news common one, which allow to subsitute users and group mention


azayati (approved):
LGTM

PR #1067, opened 2024-01-22
fix: Fix wrong displayed audience after editing published news audience - EXO-69036

sofyenne:
Prior to this change, after publishing news with a selected audience and attempting to edit it, the news audience would be incorrect. This issue was due to the uninitialized state of the news audience and its construction by the watch. This change ensures the initialization of the news audience and prevents the watch during data initialization.


azayati (approved):
LGTM

PR #1072, opened 2024-01-23
fix: Fix news illustration link accessibility - EXO-68802

sofyenne:


services/src/main/resources/locale/portlet/news/News_fr.properties
```diff
@@ -191,7 +191,7 @@ news.title.confirmDeleteDraftNews=Supprimer le brouillon?
 news.button.ok=Valider
 news.button.cancel=Annuler
 news.share.message=L'article a \u00E9t\u00E9 partag\u00E9 dans
-news.illustration.link.title=Ouvrir l'article
+news.illustration.link.title=Ouvrir l'article {0}
```
azayati, line 194: fr not needed IMO, it should be handled by crowdin
WDYT @mkrout @rdenarie ?

azayati (changes requested):

services/src/main/resources/locale/portlet/news/News_fr.properties
rdenarie: we should let it for test, it will be erased by crowdin

azayati (approved):
LGTM

PR #1089, opened 2024-02-01
fix: Article drafts deletion makes the article displayed on top of the news application - EXO-58132

sofyenne:
Prior to this change, after deleting the draft article, the article was displayed at the top of the articles list. This issue arose due to the update of the publication:liveDate property during the change of the current revision state to hide the deleted activity. This modification is going to prevent the publication:liveDate from being updated for the delete action.


azayati (approved):
LGTM

PR #1094, opened 2024-02-08
fix: fix deleted news figure in missed activities counter - EXO-69288

GouadriaHanen:
before this change, after deleting news, it appears in the missed activities counter since its metadata item still exists
After this change, when deleting news its unread metadata items are deleted and  the counter of missed activities is displayed correctly

### DIFF
--- a/services/src/main/java/org/exoplatform/news/service/impl/NewsServiceImpl.java
+++ b/services/src/main/java/org/exoplatform/news/service/impl/NewsServiceImpl.java
@@ -221,10 +221,7 @@ public class NewsServiceImpl implements NewsService {
     newsStorage.deleteNews(newsId, isDraft);
     indexingService.unindex(NewsIndexingServiceConnector.TYPE, String.valueOf(news.getId()));
     MetadataObject newsMetadataObject = new MetadataObject(NewsUtils.NEWS_METADATA_OBJECT_TYPE, newsId, null);
-    List<MetadataItem> metadataItems = metadataService.getMetadataItemsByObject(newsMetadataObject);
-    for (MetadataItem metadataItem : metadataItems) {
-      metadataService.deleteMetadataItem(metadataItem.getId(), true);
-    }
+    metadataService.deleteMetadataItemsByObject(newsMetadataObject);
     NewsUtils.broadcastEvent(NewsUtils.DELETE_NEWS, currentIdentity.getUserId(), news);
   }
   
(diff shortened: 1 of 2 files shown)


services/src/main/java/org/exoplatform/news/service/impl/NewsServiceImpl.java
```diff
@@ -214,11 +220,16 @@ public void deleteNews(String newsId, org.exoplatform.services.security.Identity
...
+    if (newsTargets != null) {
       newsTargetingService.deleteNewsTargets(newsId);
     }
     newsStorage.deleteNews(newsId, isDraft);
     indexingService.unindex(NewsIndexingServiceConnector.TYPE, String.valueOf(news.getId()));
+    MetadataObject newsMetadataObject = new MetadataObject(NewsUtils.NEWS_METADATA_OBJECT_TYPE, newsId, null);
+    List<MetadataItem> metadataItems = metadataService.getMetadataItemsByObject(newsMetadataObject);
+    for (MetadataItem metadataItem : metadataItems) {
+      metadataService.deleteMetadataItem(metadataItem.getId(), true);
+    }
```
azayati: - IMO this is already covered by newsTargetingService.deleteNewsTargets(newsId) with metadataService.deleteMetadataItemsByMetadataTypeAndObject(METADATA_TYPE.getName(), newsTargetObject); which should normally do exactly the same treatment, otherwises we should fix the problem from the metadataservice itself.
- We should keep the treatment of news targets done by newsTargetingService and not directly from newsService, should you just move this block 
`if (newsTargets != null) {
      newsTargetingService.deleteNewsTargets(newsId);
   }` here ?

azayati (changes requested):

services/src/main/java/org/exoplatform/news/service/impl/NewsServiceImpl.java
GouadriaHanen: it is not for targeting news it is for the news unread activity

azayati: ok in that case you need to delete only the metadataItems with the type "unread", otherwises you will delete all metadata items related to the newsId of all types ?

mkrout: It will be better to add a new method to the metadata service to delete all items by objectId, to avoid getting the list of items and delete them in a for loop.

mkrout: You can use metadataService.deleteMetadataItemsByObject(MetadataObject object) instead of getting the items then delete them

azayati (changes requested):
https://github.com/exoplatform/news/pull/1094#discussion_r1484530776

azayati (dismissed):
LGTM

mkrout (approved):

PR #1128, opened 2024-03-08
fix: Fix greeting text for posted/published news

hthabti:
prior to this change the greeting text for posted & published news was received in english even though the plf language of the receiver is set to another language.
after the change the greeting text would be in the adequate language


azayati (approved):
LGTM

PR #1145, opened 2024-03-20
fix: NewsList template display 2 columns of new when used in small column - EXO- 70464

rdenarie:
meeds-io/MIPs#113 update breakpoint definition
This commit adapts breakpoint usage in news template


azayati (approved):
LGTM

PR #1152, opened 2024-03-25
feat: Create the news root page if it does not exist when creating the first draft article - EXO-70331

sofyenne:


services/src/main/java/org/exoplatform/news/service/impl/NewsServiceImplV2.java
```diff
@@ -428,7 +437,16 @@ public boolean canArchiveNews(Identity currentIdentity, String newsAuthor) {
...
+    Wiki wiki = wikiService.getWikiByTypeAndOwner(WikiType.GROUP.name().toLowerCase(), pageOwnerId);
+    Page articlesRootPage = null;
+    if (wiki != null) {
+      articlesRootPage = getArticlesRootPage(pageOwnerId);
+    } else {
+      wiki = createNewsRootPage(WikiType.GROUP.name().toLowerCase(), pageOwnerId);
+      if (wiki != null) {
+        articlesRootPage = getArticlesRootPage(wiki.getOwner());
+      }
+    }
```
azayati: Here we can have already the wiki (app notes already opened), in that case the articles root page will not be retrieved.
So IMO in order to cover all cases you should:

Test if the articles root page exists, if not we test if the wiki exists, if not we create it (with the homepage), else we create just the news root page

services/src/main/java/org/exoplatform/news/service/impl/NewsServiceImplV2.java
```diff
@@ -471,7 +489,7 @@ private News createDraftArticleForNewPage(News draftArticle, String pageOwnerId,
 
   private Page getArticlesRootPage(String ownerId) {
     List<Page> notes =
-                     noteService.getNotesOfWiki("group", ownerId)
+                     noteService.getNotesOfWiki(WikiType.GROUP.name().toLowerCase(), ownerId)
```
azayati: here I think that we can retrieve the news root page directly with its name without looping

azayati (changes requested):

services/src/main/java/org/exoplatform/news/service/impl/NewsServiceImplV2.java
```diff
@@ -680,6 +703,38 @@ private Long saveArticleIllustration(String articleUploadId, Long oldArticleIllu
     }
   }
 
+
+  private Page createNewsArticlesNoteRootPage(Wiki wiki) throws WikiException {
+    if (wiki != null) {
+      Page articlesPage = new Page();
```
azayati:
```suggestion
      Page newsArticlesRootNotePage = new Page();
```

services/src/main/java/org/exoplatform/news/service/impl/NewsServiceImplV2.java
```diff
@@ -428,7 +437,21 @@ public boolean canArchiveNews(Identity currentIdentity, String newsAuthor) {
   }
 
   private News createDraftArticleForNewPage(News draftArticle, String pageOwnerId, String draftArticleCreator) throws Exception {
-    Page articlesRootPage = getArticlesRootPage(pageOwnerId);
+    Wiki wiki = wikiService.getWikiByTypeAndOwner(WikiType.GROUP.name().toLowerCase(), pageOwnerId);
+    Page articlesRootPage = null;
```
azayati:
```suggestion
    Page newsArticlesRootNotePage = null;
```

services/src/main/java/org/exoplatform/news/service/impl/NewsServiceImplV2.java
```diff
@@ -470,12 +493,12 @@ private News createDraftArticleForNewPage(News draftArticle, String pageOwnerId,
   }
 
   private Page getArticlesRootPage(String ownerId) {
```
azayati: not needed IMO, you can use getNoteOfNoteBookByName directly in the caller method

azayati (changes requested):

azayati (approved):
LGTM

PR #1160, opened 2024-03-26
feat: Create the news root page if it does not exist when creating the first draft article - EXO-70331

sofyenne:


azayati (approved):
LGTM

